Issue #10, opened 2022-03-31
Add GitLab auth provider


benjamin-schilling:
When I run typebot using the provided docker-compose file, the only login method available is GitHub OAuth, does typebot support a "username + password" login option or allows to configure a custom OAuth provider?

I want to host typebot in an company internal environment.

One additional question:
Is it possible to prevent user registration in the builder? Something like setting up a small defined set of user accounts for it only?

baptisteArno:
Check out https://app.typebot.io

By default, I allow signing up with GitHub, Google, Facebook, or Magic Link.
I prefer to avoid storing passwords. You can enable those providers by adding [dedicated environment variables](https://docs.typebot.io/self-hosting/configuration)

You can also fork the repo and add your own providers / logic. It uses [NextAuth](https://next-auth.js.org/) under the hood :)

benjamin-schilling:
I completely agree on avoiding password where possible, what is your opinion on adding support for a "generic" OAuth2/OpenID Connect provider, like a self-hosted Keycloak or similar?

baptisteArno:
Yeah, any provider would be a great thing to add! So that anyone could enable a provider by just providing the corresponding environment variable.

Any PR related to this would be greatly appreciated 🥰

Here are the instructions for Keycloak: https://next-auth.js.org/providers/keycloak

baptisteArno:
Hey,

Thank you for your work on this!

1. Then it's probably an issue on [next-auth](https://github.com/nextauthjs/next-auth) since their official Prisma adapter looks pretty much the same: https://github.com/nextauthjs/adapters/blob/main/packages/prisma/src/index.ts. It might be great to also create a pull request there! 👍 
2. Sure, sounds awesome. The goal is to be able to configure this with environment variables.

Let me know if you need any help!

benjamin-schilling:
For now we have only tested this on our side with a dev build, which works fine.

What do we have to do to run our changes in a production environment?

When I run `docker-compose up` I only get the original state deployed, our changes are missing.

Am I missing something?

baptisteArno:
If it works on dev, it should work fine in the production environment.

Feel free to create the Pull request and I'll review it 😃

baptisteArno:
Any updates on this @benjamin-schilling? 🙂

benjamin-schilling:
We are currently struggeling a little with our environment, but we will definitely create the pull request once we are convinced it's done.

PR #28, opened 2022-04-19
fix(auth): 🐛explicitly set linkAccount user values

laurin-wolf:
Only set the expected values of the user in the linkAccount method. This is necessary because GitLab includes values in the token response that are not expected.

Mentioned in [#10](https://github.com/baptisteArno/typebot.io/issues/10#issuecomment-1088907597)

### DIFF
--- a/apps/builder/pages/api/auth/adapter.ts
+++ b/apps/builder/pages/api/auth/adapter.ts
@@ -42,7 +42,22 @@ export function CustomAdapter(p: PrismaClient): Adapter {
     },
     updateUser: (data) => p.user.update({ where: { id: data.id }, data }),
     deleteUser: (id) => p.user.delete({ where: { id } }),
-    linkAccount: (data) => p.account.create({ data }) as any,
+    linkAccount: (data) => {
+      return p.account.create({
+        data: {
+          provider: data.provider,
+          type: data.type,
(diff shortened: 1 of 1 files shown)


apps/builder/pages/api/auth/adapter.ts
```diff
@@ -42,7 +42,22 @@ export function CustomAdapter(p: PrismaClient): Adapter {
...
+          type: data.type,
+          providerAccountId: data.providerAccountId,
+          access_token: data.access_token,
+          token_type: data.token_type,
+          expires_at: data.expires_at,
+          refresh_token: data.refresh_token,
+          scope: data.scope,
+          id_token: data.id_token,
+          userId: data.userId,
+        }
```
baptisteArno, line 58: What about:

```
session_state            String?
oauth_token_secret       String?
oauth_token              String?
refresh_token_expires_in Int?
```

laurin-wolf: Oh ok we went by the attributes that were explicitly specified for `Account`, but using the attributes from the prisma schema is definitely better.

laurin-wolf: Should be fixed in 43bae44eb23a80f8995776fd44261de2911935d1

Issue #217, opened 2023-01-06
Enable ctrl+z to undo changes in editor


jtapeg:
Do you think about adding the ctrl+z function?

baptisteArno:
That'd be nice indeed 👍

PR #255, opened 2023-01-15
feat(builder): Enable ctrl+z to undo changes in editor

Jorgelig:
Added shortcut Ctrl + z requested in task #[217](https://github.com/users/baptisteArno/projects/4/views/1?pane=issue&itemId=17517993)

Closes https://github.com/baptisteArno/typebot.io/issues/217

### DIFF
--- a/apps/builder/src/features/editor/components/TypebotHeader/TypebotHeader.tsx
+++ b/apps/builder/src/features/editor/components/TypebotHeader/TypebotHeader.tsx
@@ -27,6 +27,8 @@ import { headerHeight } from '../../constants'
 import { EditableEmojiOrImageIcon } from '@/components/EditableEmojiOrImageIcon'
 import { PublishButton } from '@/features/publish'
 import { CollaborationMenuButton } from '@/features/collaboration'
+import { useUndoShortcut } from '@/hooks/useUndoShortcut'
+
 
 export const TypebotHeader = () => {
   const router = useRouter()
@@ -51,6 +53,8 @@ export const TypebotHeader = () => {
     save().then()
     setRightPanel(RightPanel.PREVIEW)
   }
+
+  useUndoShortcut(canUndo, undo);
 
   const handleHelpClick = () => {
     isCloudProdInstance
--- a/apps/builder/src/features/editor/components/TypebotHeader/TypebotHeader.tsx
+++ b/apps/builder/src/features/editor/components/TypebotHeader/TypebotHeader.tsx
@@ -17,7 +17,7 @@ import {
 import { RightPanel, useEditor } from '../../providers/EditorProvider'
 import { useTypebot } from '../../providers/TypebotProvider'
 import { useRouter } from 'next/router'
-import React from 'react'
+import React, { useState } from 'react'
 import { isNotDefined } from 'utils'
 import { EditableTypebotName } from './EditableTypebotName'
 import { getBubbleActions } from 'typebot-js'
@@ -28,7 +28,7 @@ import { EditableEmojiOrImageIcon } from '@/components/EditableEmojiOrImageIcon'
 import { PublishButton } from '@/features/publish'
 import { CollaborationMenuButton } from '@/features/collaboration'
 import { useUndoShortcut } from '@/hooks/useUndoShortcut'
-
+import { useDebouncedCallback } from 'use-debounce'
 
 export const TypebotHeader = () => {
   const router = useRouter()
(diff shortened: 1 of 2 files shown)


apps/builder/src/features/editor/components/TypebotHeader/TypebotHeader.tsx
```diff
@@ -52,6 +52,39 @@ export const TypebotHeader = () => {
...
+        (event: KeyboardEvent) => {
+          if (isUndoShortcut(event) && canUndo) {
+            undo()
+          }
+        },
+        false
+      )
+    }
+  }, [canUndo, undo])
+
```
baptisteArno, line 87: Could you refactor this to a `useUndoShortcut` hook that we would use like this:

```ts
useUndoShortcut(undo)
```

baptisteArno (changes requested):

apps/builder/src/features/editor/components/TypebotHeader/TypebotHeader.tsx
```diff
@@ -52,6 +52,39 @@ export const TypebotHeader = () => {
...
   }
 
+  const isUndoShortcut = (event: KeyboardEvent) => {
+    const key = event.key.toLowerCase()
+    const cmdHeld = event.metaKey
+    const ctrlHeld = event.ctrlKey
+    const Z = 'z'
+
+    return (cmdHeld && key === Z) || (ctrlHeld && key === Z)
+  }
```
baptisteArno, line 62: It would probably be better if we don't create variables each time a key is hit. So I would suggest:

```suggestion
  const isUndoShortcut = (event: KeyboardEvent) => (event.metaKey || event.ctrlKey) && event.key === 'z'
```

apps/builder/src/features/editor/components/TypebotHeader/TypebotHeader.tsx
```diff
@@ -52,6 +52,39 @@ export const TypebotHeader = () => {
...
+        'keydown',
+        (event: KeyboardEvent) => {
+          if (isUndoShortcut(event) && canUndo) {
+            undo()
+          }
+        },
+        false
+      )
+    }
+  }, [canUndo, undo])
```
baptisteArno, line 86: You can use the `useEventListener` hook from chakra to avoid having to add a clean up function

apps/builder/src/features/editor/components/TypebotHeader/TypebotHeader.tsx
Jorgelig: added useUndoShortcut hook

apps/builder/src/features/editor/components/TypebotHeader/TypebotHeader.tsx
Jorgelig: refactored

apps/builder/src/features/editor/components/TypebotHeader/TypebotHeader.tsx
Jorgelig: I have no experience using chakra, but would it be something like that?

`  useEventListener('keydown', (event: KeyboardEvent) => {
    if (isUndoShortcut(event) && canUndo) {
      undo()
    }
  }, document, { passive: false })`

baptisteArno: Any reason why you are setting passive to `false`?

```ts
useEventListener('keydown', (event: KeyboardEvent) => {
    if (isUndoShortcut(event) && canUndo) {
      undo()
    }
  })
```

That's maybe enough

apps/builder/src/hooks/useUndoShortcut.ts
```diff
@@ -0,0 +1,14 @@
...
+    (event.metaKey || event.ctrlKey) && event.key === 'z'
+
+    useEventListener('keydown', (event: KeyboardEvent) => {
+      if (isUndoShortcut(event) && canUndo) {
+        undo()
+      }
+    }, null, { passive: false })
+}
+
+export default useUndoShortcut
```
baptisteArno, line 14: Can you not export as default? I tend to do only named exports in the project

```ts
export const useUndoShortcut =...
```

I'll add a eslint rule to enforce this 👍

baptisteArno:
I'm concerned with this scenario:

What happens if we edit a text bubble, and we just want to hit "Cmd + z" to undo the text? It would undo the whole typebot

apps/builder/src/features/editor/components/TypebotHeader/TypebotHeader.tsx
Jorgelig: refactored

apps/builder/src/hooks/useUndoShortcut.ts
Jorgelig: refactored

baptisteArno:
I'll have to test it a bit and will let you know 👍

Jorgelig:
> I'm concerned with this scenario:
> 
> What happens if we edit a text bubble, and we just want to hit "Cmd + z" to undo the text? It would undo the whole typebot

The undo functionality already exists within Typebot js. When reusing that function, it should behave the same as directly clicking the undo button in the header.

baptisteArno:
So I had to improve it a bit:

- A visual tooltip when triggering the shortcut
- Avoid triggering the shortcut when the user is typing

See https://github.com/baptisteArno/typebot.io/pull/255/commits/17b27410d70861d802669322d9f9cf7b579cba95

baptisteArno (approved):

PR #258, opened 2023-01-16
fix(builder): Calculate the completion rate

### LINKED ISSUE #244: [Results page > Analytics tab > Completion rate]: showing 'Infinity' instead of a number
It seems to happen when there's a number of views, but the number of starts is zero.

image

Jorgelig:
Calculate the completion rate if both totalCompleted and totalStarts are defined and are not zero

image

Closes #244

### DIFF
--- a/apps/builder/src/features/analytics/components/StatsCards.tsx
+++ b/apps/builder/src/features/analytics/components/StatsCards.tsx
@@ -10,6 +10,14 @@ import {
 import { Stats } from 'models'
 import React from 'react'
 
+const computeCompletionRate  = (
+  totalCompleted: number,
+  totalStarts: number
+): string => {
+  if (totalStarts === 0) return 'Not available'
+  return `${Math.round((totalCompleted / totalStarts) * 100)}%`
+}
+
 export const StatsCards = ({
   stats,
   ...props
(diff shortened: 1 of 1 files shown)


apps/builder/src/features/analytics/components/StatsCards.tsx
```diff
@@ -10,6 +10,16 @@ import {
 import { Stats } from 'models'
 import React from 'react'
 
+const calculateCompletionRate = (
+  totalCompleted: number,
+  totalStarts: number
+): string => {
+  if (!totalCompleted || !totalStarts) return 'Not available'
+  if (isNaN(totalCompleted) || isNaN(totalStarts)) return 'Not available'
```
baptisteArno, line 18: These 2 lines can be removed because we know that `totalCompleted` and `totalStarts` are defined numbers

```suggestion

```

apps/builder/src/features/analytics/components/StatsCards.tsx
```diff
@@ -10,6 +10,16 @@ import {
 import { Stats } from 'models'
 import React from 'react'
 
+const calculateCompletionRate = (
```
baptisteArno, line 13:
```suggestion
const computeCompletionRate = (
```

I prefer this naming if it's ok with you! 👌

baptisteArno (changes requested):

apps/builder/src/features/analytics/components/StatsCards.tsx
Jorgelig: done

PR #456, opened 2023-04-12
Improve translation PT

denerFernandes:
This is a very small changes, i check all texts and for me it's really good, @baptisteArno you speak portuguese? :D


baptisteArno:
Thanks mate! 🙏

PR #542, opened 2023-06-02
🐛 Fix example

dciccale:
open, close and toggle are apparently exported from @typebot.io/js package or at least that's what worked for me, I was getting unknown function error when trying to import them from @typebot/react

aslo in order for Typebot.showPreviewMessage() and Typebot.hidePreviewMessage() to work, I need to import all exports as Typebot, otherwise it didn't seem to have the named exports "showPreviewMessage" nor "hidePreviewMessage", since I was also getting an error when trying to import them.

this is my first PR here, I apologise for mistakes upfront.

It's ok if it is not meged, I just wanted to catch your attention on this potential update required in the documentation of react embed

### DIFF
--- a/packages/embeds/react/README.md
+++ b/packages/embeds/react/README.md
@@ -85,13 +85,13 @@ This code will show the bubble and let a preview message appear after 5 seconds.
 You can use these commands:
 
 ```js
-import * as Typebot from '@typebot.io/js'
+import { showPreviewMessage } from '@typebot.io/js'
 
 Typebot.showPreviewMessage()
 ```
 
 ```js
-import * as Typebot from '@typebot.io/js'
+import { hidePreviewMessage } from '@typebot.io/js'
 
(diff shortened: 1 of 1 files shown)


baptisteArno:
Thank you very much for this 🙏

> since I was also getting an error when trying to import them.

Can you tell me what error was that?

baptisteArno:
I've reverted to `import { showPreviewMessage } from '@typebot.io/js'` and `import { hidePreviewMessage } from '@typebot.io/js'` because it should work!

baptisteArno:
That'd be great if you can create an issue for `showPreviewMessage` and `hidePreviewMessage` because I can't reproduce 🙏

dciccale:
you are correct, I was able to import `{showPreviewMessage}` and `{hidePreviewMessage}` I probably overlooked what was happening, sorry

PR #619, opened 2023-07-19
Support specifying custom OAuth scope

nebulade:
For OpenID connect over OAuth, at least `openid profile` need to be set there, but since the email is also used, the `email` scope will instruct the OpenID provider to also deliver that.


nebulade:
For the Cloudron package we currently patch that file https://git.cloudron.io/cloudron/typebot-app/-/commit/d44631dca64de6822563e2fe34c203ff66c69389

baptisteArno:
Awesome, thank you!

PR #642, opened 2023-07-25
Update StarterPlanCard.tsx

jdefranco:
changed starter url from app.typebot.io to builder url in vercel : typebot-builder-teal.vercel.app


baptisteArno:
Why is this created

PR #692, opened 2023-08-14
Fix: typo in docker-compose.build.yml

dhanielo:
Fix to docker-compose.build.yml, depends_on points to an inexists service:

```
    depends_on:
      - db
```


baptisteArno:
Thank you 🙏

PR #710, opened 2023-08-24
:zap: (typebotLink) Better typebot link with merge option

### LINKED ISSUE #675: Jump function does not sequence groups
When I use the "jump" function, it stops the execution of the current flow, not continuing to the next "Groups" created, is it supposed to be like this, does the "jump" really stop when used? wanted after pulling the "jump" of the sequence to my flow

baptisteArno:
Closes #675


baptisteArno:
Rebased on main manually

PR #694, opened 2023-08-15
🌐 Split pt translation into pt (EU) and pt_BR (Brazil)

dhanielo:
- Add support to the language nuances between European Portuguese 🇪🇺🇵🇹 and Brazilian Portuguese 🇧🇷.
- Improve coverage for "pt", "pt_BR" languages to the builder editor
- Implements i18n fallback to "en" when translations are missing

PR #767, opened 2023-09-05
🌐 Split pt translation into pt (EU) and pt_BR (Brazil)

baptisteArno:
Original PR: https://github.com/baptisteArno/typebot.io/pull/694


baptisteArno:
Thank you for the great work @dhanielo !

I just added FR and DE translations and also removed the `scopedT` that you were passing as params. We should always use the `useScopedI18n` hook.

PR #79, opened 2022-06-06
Add Azure Active Directory login option

BunColak:
Hey!

I really liked the project and might use it in the future. Given many companies use Azure AD, I thought it would be helpful to add that login option as well.

- Added the option to NextAuth config.
- Modified SocialLoginButtons to include the Azure button with the SVG logo.
- Updated configuration hint to include Azure AD.
- Updated Docs.

Let me know if you require any changes :)

PR #786, opened 2023-09-07
Fix: de,fr, pt locations export

dhanielo:
Fix locations export as next-international [#79](https://github.com/QuiiBz/next-international/commit/8246ac733d5dd5e20f7d4c5956a91f73f97708e0)


baptisteArno:
Thanks I've done it in 59cc450fd36a646230b31c431ab6b33586aa158b 🙏

PR #904, opened 2023-10-09
#FIX without Performance issue: Format email input to lowercase

Raju-kadel-27:
### Summary by CodeRabbit

- Bug Fix: Updated the `SignInForm` component in the authentication module. User email input is now automatically converted to lowercase during the sign-in process. This change enhances consistency and prevents potential issues related to case sensitivity when handling email addresses.


baptisteArno:
Hey, thank you. Though, it seems like this is not an issue, Nextauth already normalizes the email entered!

Issue #567, opened 2023-06-19
ctrl + mouse scroll zooms a lot


NicolasJardin:
By turning the mouse wheel just once the zoom is drastically increased or decreased.
Is there any way to reduce the amount of zoom per scroll to leave a zoom similar to the + and - buttons?

[2f80416a-9383-4e54-a3c8-4a762b61b05c.webm](https://github.com/baptisteArno/typebot.io/assets/66677460/3fadb251-7d35-4ed3-9433-bd060144fdb6)

evoxf1:
> /attempt #567
> Options

sorry brother, u can take this, i took another, i tried resigning but it wont work

PR #940, opened 2023-10-16
Fixed pinch zooming mouse issue(with ctrl key)

prateekkalra:
**Fixed the drastic zoom increase decrease on ctrl + mouse scroll.**

The issue was occuring due to the "scale" property in the pinch gesture. The scale was getting bigger values, leading to more zooming.
What I did was, made sure that maximum scale difference between current and last value cannot be more than the scaling factor used in zoomin/zoomout buttons. That is. 0.2
Also, the pinch zoom would work as expected.

/claim #567

### DIFF
--- a/apps/builder/src/features/graph/components/Graph.tsx
+++ b/apps/builder/src/features/graph/components/Graph.tsx
@@ -171,7 +171,12 @@ export const Graph = ({
   }) => {
     const { x: mouseX, y } = mousePosition ?? getCenterOfGraph()
     const mouseY = y - headerHeight
-    let newScale = scale ?? graphPosition.scale + (delta ?? 0)
+    let newScale = graphPosition.scale + (delta ?? 0)
+    if (scale) {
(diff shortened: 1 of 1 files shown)


baptisteArno:
Did you manage to reproduce the issue stated?

I'm kind of skeptical to merge this because I can't reproduce the issue even with my mouse.

baptisteArno:
Why is the issue referring to `Ctrl` key? Have you tried it as well?

prateekkalra:
@baptisteArno 
Sorry if I was unclear above. What I just mentioned above, was reproduced when I scrolled my mouse while pressing the Ctrl key. 
So that's the issue. Ctrl+mouse scrolling leads to inconsistent scaling.

prateekkalra:
@baptisteArno Did you make your mind yet?

baptisteArno:
Thank you so much for this 🙌

Issue #458, opened 2023-04-13
Option to set bubble height and width


lpadula:
Hi! It would be great to be able to set bubble chat height and width.

baptisteArno:
Are you referring to this?

![CleanShot 2023-04-14 at 09 47 04@2x](https://user-images.githubusercontent.com/16015833/231979158-a18fd564-fa62-46b6-a0d7-c10378dddde2.png)

lpadula:
Yes!

satoshinotdead:
Related: I was need to hard-edit the JS file in order to prevent that bubble button width and height 'scale after click' and not in loading.

My knowledge coding is very little but BTW I hope it helps :)

baptisteArno:
> to prevent that bubble button width and height 'scale after click' and not in loading.

Can you explain? I'm not sure to understand

lpadula:
He edited a js file to achieve that. I think that the user should be able to add some parameters on the installation script or maybe directly on the "theme" options.

rishi-raj-jain:
@baptisteArno

Is it open to new attempts?

mithleshfantezie:
/attempt https://github.com/baptisteArno/typebot.io/issues/458

PR #945, opened 2023-10-17
Made bubble height width configurable through props.

prateekkalra:
/claim #458
Made bubble height width configurable through props.

Please find default height width screenshot below.

image

### Summary by CodeRabbit

- New Feature: Introduced customizable dimensions for chat bubbles. Users can now specify the height and width of the chat bubbles, providing greater flexibility in adapting the chat interface to different layouts and designs. If no dimensions are specified, the system will use default values, ensuring a consistent user experience. This update enhances the visual adaptability of our chat feature, making it more user-friendly and versatile for various display settings.

### DIFF
--- a/apps/builder/src/features/blocks/integrations/webhook/helpers/getDeepKeys.ts
+++ b/apps/builder/src/features/blocks/integrations/webhook/helpers/getDeepKeys.ts
@@ -42,7 +42,9 @@ export const getDeepKeys = (obj: any): string[] => {
 
 const parseKey = (key: string) => {
   if (
-    (key.includes(' ') || key.includes('-')) &&
+    (key.includes(' ') ||
+      key.includes('-') ||
+      !isNaN(key as unknown as number)) &&
     !key.includes('.flatMap(item => item') &&
     !key.includes("['") &&
     !key.includes("']")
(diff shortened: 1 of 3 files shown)


baptisteArno (changes requested):
Here is what I have as default size:

CleanShot 2023-10-18 at 15 13 42@2x

prateekkalra:
The difference in images might be due to screen resolution. 
The default width should be same as what it is right now, in the live version ? Is it not ?

baptisteArno:
> The difference in images might be due to screen resolution. The default width should be same as what it is right now, in the live version ? Is it not ?

Width is ok in my screenshot, it's the height that is not! Please checkout the main branch to check the default styling 🙏

prateekkalra:
Sure @baptisteArno 
Will fix it.Thanks 🙂

prateekkalra:
Hi @baptisteArno 
I checked again, the height in both the main branch and my branch looks exactly similar.
Did you check out this branch and test?

baptisteArno:
It should look like this by default:

CleanShot 2023-10-20 at 07 53 00@2x

prateekkalra:
@baptisteArno You were right about the height issue. It was looking same for me, because of the smaller resolution in my laptop. Pushed a fix to align the default height with main branch. Removed max-height, since it's already been taken care by the tailwind class.
Let me know if you still have any issues.
Thank you 🙂

baptisteArno:
Sorry the issue lacked some context and clear scope. I'll close this as it does not fully satisfy what I was thinking about 🙏

Issue #352, opened 2023-03-02
Highlight JS code errors in `CodeEditor`


erickriva:
Variables injection is working fine, but the code itself is not executed.

![image](https://user-images.githubusercontent.com/2746558/222489692-84304d2d-13a5-4725-9d8c-eb5fa58bfd19.png)
![image](https://user-images.githubusercontent.com/2746558/222490057-e73ee7b2-b53a-43b7-aae7-1a74ad622194.png)
![image](https://user-images.githubusercontent.com/2746558/222490671-afc82d43-1217-4515-ae32-456ba6cdf17a.png)

erickriva:
It was an error in my code, an additional `)` in `foundIndex` declaration 🤡.
There is no way to warn this type of errors inside the block?

b4s36t4:
@baptisteArno After some research and googling I found these 3 plausible solutions

1. By default codemirror doesn't have any linting available (ex: codepen.io) which doesn't throw any error. But codemirror does provide a linter package which we can leverage to parse the tree and return any diagnostic based on tree-parse.

2. As we're using `javascript` language extension provided by codemirror itself which does have `eslint` support embedded into it.

> We might need to work with eslint configuration for browser as eslint doesn't have built-in support for browser, although there's an example provided for that.

3. Instead of using codemirror, why can't we move to monaco-editor https://github.com/microsoft/monaco-editor. I have used monaco-editor before and it's seamless. 

> Monaco might take some-time as it's an entire removal of a package and replacing with new one, maybe we can have it as a new feature or so?

Attaching the implementation of the 1st option as mentioned above.

<img width="431" alt="Screenshot 2023-10-16 at 11 57 48 PM" src="https://github.com/baptisteArno/typebot.io/assets/59088937/b814bad8-21c9-4613-9d1a-4459cc15a969">

b4s36t4:
Another solution is to running linters in wasm and use that to provide diagnostic.

example: https://quick-lint-js.com/ https://quick-lint-js.com/demo/

like we have boimejs which does have wasm lib to provide linting.

b4s36t4:
https://github.com/biomejs/biome/blob/main/website/src/playground/CodeMirror.tsx Here is full working example for integrating biome with browser env.

baptisteArno:
Thank you for the exploration @b4s36t4. I guess option 1 would be enough? The idea is to only catch syntax issues.

If you hover on it, does it display a helper tooltip?

[...]

PR #956, opened 2023-10-22
Added JS code errors in script block.

prateekkalra:
/claim #352

Added feature to show JS code errors in the script block like an IDE.
This PR shows the exact error unlike the other one which shows a static message. Also the error tooltip shows up in a single line over the script block popup. (The css changes in the PR are for adjusting the tooltip over the popup, which was a bit tricky)

Please find the demo video of the script block below.

https://github.com/baptisteArno/typebot.io/assets/29385192/5601223b-2e5c-4615-81b3-570d4397c4c3

### Summary by CodeRabbit

- New Feature: Integrated a JavaScript linter into the code editor, enhancing code quality by enforcing coding standards and highlighting potential issues in real-time.
- Improvement: Adjusted the maximum height of the code editor for better visibility and user experience.
- Improvement: Implemented sticky positioning for the code editor container, ensuring it remains visible while scrolling.
- New Feature: Added conditional padding and display settings to the PopoverContent and PopoverBody components, improving the visual presentation based on the block type. 

These changes aim to improve the user experience by enhancing code readability, enforcing coding standards, and improving the visual presentation of components.


baptisteArno:
That's very nice.

I have several questions though.

- Why did you change the max-height of the editor?
- Why did you add padding only for script block this seems like a hack. What happens if you remove this padding?
- Does this work with CSS editor (Theme tab)?
- Does this work with JSON editor (Webhook block > Advanced config > custom body)?
- Does this work with Set variable block?

prateekkalra:
@baptisteArno 
1. I changed the max height of editor since there was an issue when the code in the editor was very long.(Showed 2 scrollbars, and also didn't look good on smaller screens.) (Set editor max height to 400px which is the max height of the block)
2. Had to set the padding because in order to achieve a popup extend another popup, the display property of the block needed to be changed, which removed the padding. Now, to achieve this popup over popup, either I could get away with this hack, or I would have to change lot of parent classes.(Not sure if it's even possible then). (It only affects the script block)
3., 4., 5. The css editor, json editor and set variable block work as before. There won't be any change since the scope of this issue was highlighting JS code errors in the code editor.

baptisteArno:
Thank you for the clarification!

It seems it does not work properly with Typebot's variables templating system:

This can be a valid script:

CleanShot 2023-10-25 at 18 15 34@2x

baptisteArno:
> It seems it does not work properly with Typebot's variables templating system:

@prateekkalra Did you have the time to check this?

prateekkalra:
@baptisteArno I checked it. Sorry, there's no straightforward way to whitelist template variables as valid syntax. I understand, why this can be a dealbreaker 😞

baptisteArno:
Indeed, after investigating this myself, I don't think there's a clear solution for this... Still, thank you for your work 🙏

Issue #575, opened 2023-06-24
Editor performance issues


baptisteArno:
https://github.com/baptisteArno/typebot.io/assets/16015833/55f6ec72-ca15-49d1-87af-6ec47a8f50e1

baptisteArno:
Also on a large flow when trying to focus group or edge, it's laggy.

Here is a flow example that causes a performance problem:

```json
[...]

PR #959, opened 2023-10-22
fix graph flickering & performance improvements

neo773:
This PR fixes the flickering and improves the performance so panning around the graph is much smoother than before.

https://github.com/baptisteArno/typebot.io/assets/62795688/56b91e20-1eb0-44b5-9a4a-c07525c2ba48

/claim #575

### Summary by CodeRabbit

- Refactor: Improved the Graph component's scaling calculation for enhanced readability and maintenance.
- Style: Updated the Graph component's style properties to ensure better compatibility and visual performance on webkit browsers.

These changes aim to enhance the user experience by ensuring the Graph component displays consistently across different web browsers. The refactoring of the scaling calculation also makes the code easier to understand and maintain, potentially leading to quicker updates and bug fixes in the future.

### DIFF
--- a/apps/builder/src/features/graph/components/Graph.tsx
+++ b/apps/builder/src/features/graph/components/Graph.tsx
@@ -174,7 +174,9 @@ export const Graph = ({
     let newScale = graphPosition.scale + (delta ?? 0)
     if (scale) {
       const scaleDiff = scale - graphPosition.scale
-      newScale += Math.min(zoomButtonsScaleBlock, Math.abs(scaleDiff)) * Math.sign(scaleDiff)
+      newScale +=
+        Math.min(zoomButtonsScaleBlock, Math.abs(scaleDiff)) *
+        Math.sign(scaleDiff)
     }
 
     if (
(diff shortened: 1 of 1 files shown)


baptisteArno:
Hey @neo773 that seems to do the trick 🙌

I'm so sorry but I thought the bounty included this comment from the issue (https://github.com/baptisteArno/typebot.io/issues/575#issuecomment-1605502689). While setting the bounty price, I thought it included that comment as well which should be harder to fix. Really sorry for the confusing spec. I had to update the bounty price 🙏

baptisteArno:
Can you explain what all these CSS props do? And why are you using the webkit prefix?

neo773:
@baptisteArno 
Chromium's rendering engine was based on Webkit hence the vendor prefix
This code adjust the Z plane, the flickering was caused by Z plane not being constant.

>I thought it included that comment as well which should be harder to fix

I just saw it, it was hidden comment.

baptisteArno:
> @baptisteArno Chromium's rendering engine was based on Webkit hence the vendor prefix This code adjust the Z plane, the flickering was caused by Z plane not being constant.
> 
> > I thought it included that comment as well which should be harder to fix
> 
> I just saw it, it was hidden comment.

Would `'-webkit-perspective': 1000` be enough then?

neo773:
@baptisteArno I've found the other properties boost the performance by a bit.

PR #1114, opened 2023-12-21
Input blocks pt-BR translation

gabrielgpavao:
## Summary by CodeRabbit

- **New Features**
  - Integrated localization support across various components using the `useTranslate` hook for dynamic translations.

- **Enhancements**
  - Replaced hardcoded text with localized strings to support multiple languages in the user interface.

- **User Interface**
  - Updated labels, placeholders, tooltips, and button texts to utilize translated content for a multilingual experience.


baptisteArno:
I had to remove all the default values as it can't be internationalized for now. Also removed the Dropdown dynamic key parsing (for Condition for example) as it should be done differently. Keys shoud not be built dynamically.

Issue #627, opened 2023-07-23
Add new encryption option for a variable


baptisteArno:

klinux:
Hey @baptisteArno — we need this feature in production at my company, so I started an implementation and would love your feedback before investing in UI work.

**Draft PR:** #2458

**TL;DR design:** a new workspace-scoped `WorkspaceSecret` entity (separate from `Variable`), encrypted at rest with the same AES-256-GCM pattern as `Credentials`. Referenced from typebots via `{{$secrets.NAME}}` and resolved **only** inside explicitly marked server-side block executors (this first PR wires HTTP request block: URL, headers, body, basic auth). Secrets are never written to `Result.variables`, `SetVariableHistoryItem`, chat replies, or any client-visible payload — they're fetched, substituted, and discarded just-in-time.

**Why a separate entity instead of an `isSecret` flag on `Variable`:** `Variable` flows through ~35 call sites including text bubbles, client responses, code blocks executed in the browser, etc. A runtime flag is fragile — one missed check leaks the secret. A distinct type makes mixing a compile error.

I split the work into 3 PRs to keep reviews small: (1) schema + backend + HTTP block integration [this one, drafted], (2) builder UI (Workspace Settings → Secrets + autocomplete + save-time validation), (3) docs. **I'm holding #2 and #3 until you validate the design in #2458** so we don't waste effort if you want a different approach.

Main questions I'd love your read on, summarized in the PR:

1. Separate entity vs. `isSecret` flag on `Variable` — I went separate for safety, but happy to pivot.
2. `{{$secrets.NAME}}` syntax — alternatives welcome.
3. Should I extend server-side resolution to email/Stripe/forge blocks in this PR, or keep it HTTP-only and follow up?
4. Want it gated behind a feature flag so it can merge incrementally?
5. Migration file — generate myself or leave for you?

Happy to rework anything. Thanks for building this project! 🙏

klinux:
Hey @baptisteArno — following up since #2458 was closed as stale without discussion.

No worries on the timing, but before I invest more time I'd like to know whether this feature is something you'd want in typebot at all. A quick signal works:

- 👍 **interested** — I'll rework based on whatever direction you prefer (separate entity, flag on `Variable`, different syntax, smaller scope, etc.) and reopen.
- 👎 **not a priority / out of scope** — totally fine, feel free to close this issue and I'll maintain it in a fork.

Either answer is helpful. Thanks!

baptisteArno:
[...]

PR #1142, opened 2024-01-10
QRcode block

dkashikar:
This block is able to :

1. Show a QrCode
2. Optionally link to an external url

## Summary by CodeRabbit

- **New Features**
	- Introduced QR code creation and display capabilities in the app.
	- Added a new component for rendering QR code images.


baptisteArno:
Hey!

That's a nice idea 👍

The `qrcode-generator` package (https://github.com/kazuhikoarase/qrcode-generator#readme) was not update since 3 years. I'd prefer we use the most recent and popular package: https://github.com/soldair/node-qrcode

baptisteArno:
I also think it would be better to generate the QR Code as an image URL that can be then used in an Image bubble. So that we can use this block on other environment like whatsapp

dkashikar:
Thanks. 

> Hey!
> 
> That's a nice idea 👍
> 
> The `qrcode-generator` package (https://github.com/kazuhikoarase/qrcode-generator#readme) was not update since 3 years. I'd prefer we use the most recent and popular package: https://github.com/soldair/node-qrcode

i can rewrite using this package

> I also think it would be better to generate the QR Code as an image URL that can be then used in an Image bubble. So that we can use this block on other environment like whatsapp

super. have done this.

baptisteArno:
@dkashikar I've renamed things to proper camel case when necessary and I also properly set up the image generation so that it is uploaded to S3 instance. Otherwise, a data URI can't be used in WhatsApp.

Thanks again for your work!

baptisteArno:
Thanks again for your work @dkashikar. 🙏

Let me know if there's anything we can improve in terms of DX.

dkashikar:
Thank you @baptisteArno 
I was also thinking of having a password kind of input field. Basically masking the input with *. 
I am planning to use the bot on kiosks. So having some input hidden would be a good idea.

baptisteArno:
> Thank you @baptisteArno I was also thinking of having a password kind of input field. Basically masking the input with *. I am planning to use the bot on kiosks. So having some input hidden would be a good idea.

That would be nice, we need a system to encrypt the value then because I don't want to make the user think his value is private when it's not in reality. I'll work on that ASAP (#627)

PR #1152, opened 2024-01-12
Update overview.mdx - add cal.com incompatible with whatsapp

altomarketing:
add cal.com incompatible with whatsapp


baptisteArno:
Cheers!

PR #1181, opened 2024-01-23
OpenAI function calling fixes and improvements

vforvilela:
## Summary by CodeRabbit

- **New Features**
	- Enhanced chat functionality with improved processing of tool calls and execution flow.
- **Refactor**
	- Updated the logic for handling outputs and errors in function executions to ensure more accurate type checks and transformations.


packages/forge/blocks/openai/actions/createChatCompletion.tsx
```diff
@@ -194,33 +197,44 @@ export const createChatCompletion = createAction({
...
         messages.push(message)
 
         for (const toolCall of message.tool_calls) {
           const name = toolCall.function?.name
           if (!name) continue
           const toolDefinition = options.tools?.find((t) => t.name === name)
           if (!toolDefinition?.code || !toolDefinition.parameters) continue
+          const toolParams = Object.fromEntries(
+            toolDefinition.parameters.map(({ name }) => [name, null])
+          )
```
baptisteArno, line 230: Can you explain why you are adding this?

vforvilela: If the LLM decides not to use an optional parameter, we'll encounter an error because the variable will not be defined. The objective of this code is to ensure that all variables are defined (and set to null by default).

vforvilela:
> I can't push a commit to that PR, do you know why?
You're trying to push it to my project.

PR #1241, opened 2024-02-12
fix openapi document filepaths

rpmccarter:
👋 Hi! Engineer from mintlify here

This updates the `openapi` field in `mint.json` to point to the correct files in the repo. This bug seems to have been introduced [during the migration](https://github.com/baptisteArno/typebot.io/pull/1115/files#diff-f9f0cfdf7078569b552a4f90df1b3d7982ef914aebf74ea7d291604e91fdc1a3)


baptisteArno:
Wow indeed! Thank you for the fix 🙏

PR #1296, opened 2024-02-27
recieve from env the number off group size to sent to whatsapp API

thercd:
When using the whatsapp oficial api or another that use the same schema, with env we can modify the number of groups that the interactive group will be split.


baptisteArno:
Why would you need that? The maximum is 3, this won't be possible to increase that number

thercd:
> Why would you need that? The maximum is 3, this won't be possible to increase that number

When using another API, changing the env WHATSAPP_CLOUD_API_URL to another URL, another APIs can have different implementations and accept a higher number.

baptisteArno:
> > Why would you need that? The maximum is 3, this won't be possible to increase that number
> 
> When using another API, changing the env WHATSAPP_CLOUD_API_URL to another URL, another APIs can have different implementations and accept a higher number.

I doubt that because the limitation is on WhatsApp side. Can you provide a screenshot of more than 3 buttons on the WhatsApp app?

thercd:
sure, actually the buttons are converted to lists in this custom API  (link if you want do take a look https://github.com/clairton/unoapi-cloud), so lists can handle with many items, in the first exemple the env is set to handle 100 items so the whatsapp dont split the list, in the second one, is was the value set to 2, so 3 lists are created.

exemple1


exemple2
exemple2 1

apps/docs/self-hosting/configuration.mdx
```diff
@@ -245,6 +245,7 @@ In order to be able to test your bot on WhatsApp from the Preview drawer, you ne
 | WHATSAPP_PREVIEW_TEMPLATE_NAME        |                            | The preview start template message name                 |
 | WHATSAPP_PREVIEW_TEMPLATE_LANG        | en_US                      | The preview start template message name                 |
 | WHATSAPP_CLOUD_API_URL                | https://graph.facebook.com | The WhatsApp Cloud API base URL                         |
+| WHATSAPP_INTERACTIVE_GROUP_SIZE       | 3                          | The array size off itens to send to API on choice input |
```
baptisteArno, line 248:
```suggestion
| WHATSAPP_INTERACTIVE_GROUP_SIZE       | 3                          | The array size of items to send to API on choice input. You can't choose a number higher than 3 if you are using the official cloud API URL.  |
```

baptisteArno (changes requested):

baptisteArno:
Ok interesting, thank you for the info :)

baptisteArno (approved):

Issue #1321, opened 2024-03-05
Preview Message button's accessibility issue


yaditya:
Typebots’ **Close** button has been picked up as a critical accessibility issue according to [aXe DevTools extension](https://chromewebstore.google.com/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd) with the exact issue being **Buttons must have discernible text** (see the screen capture below). Upon further digging, the issue lies on this [particular section](https://github.com/baptisteArno/typebot.io/blob/275ba1b1c40f4e79d42de7915e02fbe5ed1955b5/packages/embeds/js/src/features/bubble/components/PreviewMessage.tsx#L68) where the `<button class="preview-message-close-button">` element should ideally have `aria-label` for screen readers to be able to understand what that button does.

![typebot preview message close button](https://github.com/baptisteArno/typebot.io/assets/1902315/641c4490-2210-4131-b1a1-622d566c6e96)

PR #1325, opened 2024-03-05
Fix Issue #1321 Preview Message button's accessibility issue

KavitaR45:
Added aria-label for the button <button class="preview-message-close-button">

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,1990 @@
+{
+  "name": "typebot-os",
+  "version": "2.23.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "typebot-os",
+      "version": "2.23.0",
+      "license": "AGPL-3.0-or-later",
+      "devDependencies": {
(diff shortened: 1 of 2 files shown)


package-lock.json
baptisteArno, line 1: What's this?

packages/embeds/js/src/features/bubble/components/PreviewMessage.tsx
```diff
@@ -67,6 +67,7 @@ const CloseButton = (props: {
 }) => (
   <button
     part="preview-message-close-button"
+    aria-label="preview-message-close-button"
```
baptisteArno, line 70:
```suggestion
    aria-label="Close"
```

it should act like a real button label :)

baptisteArno (changes requested):

package-lock.json
KavitaR45: Ohh thanks for the suggestion, I thought close would be more common like other button would be using it. Thanks anyway.

baptisteArno (changes requested):
Why is this PR adding a `package-lock.json` file?

PR #1339, opened 2024-03-09
New Feature : Password Protection

MatthewDlr:
Hey all,
I've recently self-hosted typebot for a school project, and I've developed a new feature that allows a bot to be protected by a password. When someone accesses a bot that has a password set up, he will have to enter the password in order to start the chat. 

It's my very first time on Next.js (and react) so my changes may need improvements ; feel free to suggest me modifications, if necessary. 

Oh, and I've also translated the “share” page to French. 

Looking forward to seeing these changes implemented 🚀👀.

CleanShot 2024-03-09 at 18 03 20
CleanShot 2024-03-09 at 18 02 03


apps/viewer/src/components/TypebotPageV3.tsx
```diff
@@ -54,11 +58,89 @@ export const TypebotPageV3 = ({
...
     >
       <SEO url={url} typebotName={name} metadata={metadata} />
-      <Standard
-        typebot={publicId}
-        onInit={clearQueryParamsIfNecessary}
-        font={font ?? undefined}
-      />
+
+      {isPasswordProtected || !isPasswordValid ? (
+        <div
```
MatthewDlr, line 63: Since Chakra UI is not available here, ended up doing regular html css to make this form

apps/viewer/src/components/TypebotPageV3.tsx
```diff
@@ -54,11 +58,89 @@ export const TypebotPageV3 = ({
       }}
     >
       <SEO url={url} typebotName={name} metadata={metadata} />
-      <Standard
-        typebot={publicId}
-        onInit={clearQueryParamsIfNecessary}
-        font={font ?? undefined}
-      />
+
+      {isPasswordProtected || !isPasswordValid ? (
```
MatthewDlr, line 62: Since Chakra UI is not available here, ended up doing regular html css to make this form

MatthewDlr:
Thank you so much for your feedback.

Sorry for not having talked about it before; I just developed that for my association and afterward, I thought that it would be a good idea to submit my change. For the future I’ll create a discussion.
About moving the settings UI and the protection logic, I see what you mean it’s ok; but to change the password UI in a bot message I don’t see how to achieve that. Besides that, if the first message is a password, which would mean that the conversation has started, and the bot is loaded; but if so, the password would be visible in the network console. Can you provide me with some cues on that please so I can move forward faster.
Thanks!

baptisteArno:
I think we will close this in favor of #627 🙏

Thank you for working on this still ❤️

PR #1341, opened 2024-03-11
Button Dynamic Data to have more advanced possibilities 

abdullahbaa5:
Button choice dynamic data to allow different display name vs value set in variable

sometimes we want to save key of the choice rather than it's name (due to localization in multiple languages and many other use cases as well)

Normal Usecase:
image

Advanced Usecase:
image

I have added a few steps of validation to make sure in case of json data but no id, name we use the whole json array but you can define the standard key which are (id and name) yourself or give option in the button pop.

Changes required:
- Add missing type definitions (required) - will not be able to build docker file without types (currently im bypassing type check on my local workspace)
- Use built in type bot functions to check whether the variable value is string, json (optional)
- Ability to change the variable keys for displaying and variable saving (optional)
- Change how and where the display name and variable id (reply, replyid) are present (optional)

Tested with different data types for every scenario: yes and works as expected.

### DIFF
--- a/packages/bot-engine/blocks/inputs/buttons/injectVariableValuesInButtonsInputBlock.ts
+++ b/packages/bot-engine/blocks/inputs/buttons/injectVariableValuesInButtonsInputBlock.ts
@@ -38,7 +38,12 @@ export const injectVariableValuesInButtonsInputBlock =
             contentAsJson = false;
           }
 
-          if (contentAsJson)
+          if (contentAsJson
+            &&
+            (contentAsJson.name && contentAsJson.name !== "") 
+            &&
+            (contentAsJson.id && contentAsJson.id != "")
+          )
           {
             result = Object.assign({}, result, {
               content: contentAsJson.name,
--- a/packages/bot-engine/blocks/inputs/buttons/injectVariableValuesInButtonsInputBlock.ts
+++ b/packages/bot-engine/blocks/inputs/buttons/injectVariableValuesInButtonsInputBlock.ts
@@ -23,11 +23,37 @@ export const injectVariableValuesInButtonsInputBlock =
       const value = getVariableValue(state)(variable)
       return {
         ...deepParseVariables(variables)(block),
-        items: value.filter(isDefined).map((item, idx) => ({
-          id: idx.toString(),
-          blockId: block.id,
-          content: item,
-        })),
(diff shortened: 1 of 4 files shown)


baptisteArno:
On Typebot, the user can only save strings or list of strings. We don't support JSON and this feature implicitely uses JSON which makes it then confusing.

There is already an example on how to achieve what you're doing with the [Map with same index](https://docs.typebot.io/editor/blocks/logic/set-variable#map-item-with-same-index) set variable value.

I'm sorry for closing this but you should have created an issue before to discuss it 🙏

abdullahbaa5:
> On Typebot, the user can only save strings or list of strings. We don't support JSON and this feature implicitely uses JSON which makes it then confusing.
> 
> There is already an example on how to achieve what you're doing with the [Map with same index](https://docs.typebot.io/editor/blocks/logic/set-variable#map-item-with-same-index) set variable value.
> 
> I'm sorry for closing this but you should have created an issue before to discuss it 🙏

unlike the picture choice, the button choice returns the text rather than the index which prompt me to use json for button choice but i think the functionality to return array index for the button can be implemented rather than just displaying the value of the button itself.

baptisteArno:
What do you mean by the picture choice does not return the index? That's not true, it returns either the description or the image URL.

I don't like the idea to implicitely parse a JSON variable. It sends the message that saving JSON in a variable is OK on Typebot, which is not.

Something that I'd be ok with is to have a second variable selector for the matching list of IDs. So you'd be able to provide both: list of labels and list of ids and it would then automatically saves the matching ID.

abdullahbaa5:
> What do you mean by the picture choice does not return the index? That's not true, it returns either the description or the image URL.
> 
> I don't like the idea to implicitely parse a JSON variable. It sends the message that saving JSON in a variable is OK on Typebot, which is not.
> 
> Something that I'd be ok with is to have a second variable selector for the matching list of IDs. So you'd be able to provide both: list of labels and list of ids and it would then automatically saves the matching ID.

great! that would make more sense as the functionality i am here trying to achieve is to separate the display name from the ids. Let me see what I can do.

baptisteArno:
In terms of UI, it can be like the picture choice settings:

CleanShot 2024-03-11 at 09 12 02

"Labels", "Values"

PR #1343, opened 2024-03-11
Button dynamic data with value and labels

abdullahbaa5:
Added an extra field for values (ids) + names (labels) for button choices.

It is backward compatible (will always try to use one or the other in case one is missing so users who have previously set values only do not lose any functionality)


baptisteArno:
I'm sorry, after thinking about it, I'd prefer not to implement that.

I think it will be confusing for new users to be able to set 2 different variables.

As for users who'd like to achieve this, adding a Set variable block with [Map item with same index](https://docs.typebot.io/editor/blocks/logic/set-variable#map-item-with-same-index) right after the Buttons block is really easy to implement.

abdullahbaa5:
> I'm sorry, after thinking about it, I'd prefer not to implement that.
> 
> I think it will be confusing for new users to be able to set 2 different variables.
> 
> As for users who'd like to achieve this, adding a Set variable block with [Map item with same index](https://docs.typebot.io/editor/blocks/logic/set-variable#map-item-with-same-index) right after the Buttons block is really easy to implement.

the issue with mapping is that it assumes that there are not multiple items of same name, same issue exists with the button feature.

abdullahbaa5:
is it possible to create a block similar to choices (that will display buttons) using the forge? this even if you decide not to go down this road, atleast I can/will be able to use it for my usecase while staying up-to date with the typebot latest updates without having to fix conflictions every update :D

abdullahbaa5:
I have found a better alternative, a checkbox (return index) and this will return the index of the selected item rather than the value of the button.

This will tackle the issue of not adding any extra field for dynamic data.

PR #1351, opened 2024-03-13
Checkbox in Button/Picture Choices to return the position of the choice

abdullahbaa5:
A non-conflicting addition without introducing any extra variable addition for button choices and picture choices block with a checkbox to save position of the choice chosen rather than the text (or other value) of the choice.


baptisteArno:
I don't understand the point of this? What's the use case?

abdullahbaa5:
> le, provide a real case scenario

it can but like i said, if there are choices of the same name, it will cause issues.

Use case: real estate buildings.

example choices:
(Flat Newyork 5bhk) (Flat Newyork 5bhk) (Flat Newyork 4bhk) now if user chooses second option, mapping will never get it right. Even though the naming are same (the picture and description are different).

baptisteArno:
So then it will not work with the current engine as the payload sent to the server would be:

```ts
{
  "message": "Flat Newyork 5bhk"
}
```

It will never know which one was selected. And this payload can't be change as it also need to be compatible with WhatsApp runtime. 

This is indeed problematic.

baptisteArno:
`matchedItem.id` is not the index, it's the typebot item ID.

abdullahbaa5:
> WhatsApp

oh right... the only reason i am creating the bots are for whatsapp so i guess this would have to be scrapped... i will just manually assign a unique id to everything in the list

abdullahbaa5:
> `matchedItem.id` is not the index, it's the typebot item ID.

I understand but it is set as  the position of the choice so i thought there is no extra processing required

abdullahbaa5:
it seems the whatsapp api is able to return a payload or some kind of id related to the buttons that were created, maybe we could use that to fix the whatsapp issue?

baptisteArno:
I think the real problem is displaying non-unique names in the Buttons input. That does not really make sense for the end user. Why would he select the first "Flat Newyork 5bhk" over the third?

abdullahbaa5:
> I think the real problem is displaying non-unique names in the Buttons input. That does not really make sense for the end user. Why would he select the first "Flat Newyork 5bhk" over the third?

understandable... maybe the foundation needs to be changed where one needs not to duplicates. I think we can close this request!

Issue #997, opened 2023-11-02
Integrate with NocoDB


CptPlastic:
I would love to have this in addition to google sheets. I have a NocoDB instance for testing as needed. I can send invites as needed

https://noco.app.p7n.net/

Issue #970, opened 2023-10-26
Integrate with NocoDB and Baserow


jdaviescoates:
It'd be nice if responses could be stored in open source airtable-like apps such as [NocoDB](https://github.com/nocodb/nocodb) and [Baserow](https://gitlab.com/baserow/baserow). Would much prefer that than e.g. Google Sheets (as convenient as that is for many people)

baptisteArno:
#997

PR #1365, opened 2024-03-15
NocoDB Integration

abdullahbaa5:
#970 #997

Fully integrate NocoDB.

Added all API Functions:

- List Table Records
- Create Table Records
- Update Table Records
- Delete Table Records
- Read Table Record
- Count Table Records
- List Linked Records
- Link Records
- Unlink Records

Optional Todo:
- Save responses of non-get requests in a variable (error validation try-catch is added and logged so i do not think so it is much needed)

You are free to implement any extra validation/function :D


packages/forge/blocks/nocodb/actions/deleteTableRecords.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const deleteTableRecords = createAction({
+  auth,
+  name: 'Delete Table Records',
+  options: option.object({
+    tableId: option.string.layout({
+      label: 'Table ID',
+      isRequired: true,
+      helperText: 'Identifier of the table to delete records from.',
+    }),
```
baptisteArno, line 14: Can this be a dynamic select field like the OpenAI model option?

packages/forge/blocks/nocodb/actions/listTableRecords.ts
```diff
@@ -0,0 +1,105 @@
...
+import { isDefined } from '@typebot.io/lib'
+
+export const listTableRecords = createAction({
+  auth,
+  name: 'List Table Records',
+  options: option.object({
+    tableId: option.string.layout({
+      label: 'Table ID',
+      isRequired: true,
+      helperText: 'Identifier of the table to retrieve records from.',
```
baptisteArno, line 14: I'm not a fan of having a helper text for each input.

Here is the protocol I follow:
- use 'moreInfoTooltip' to provide more info if the label is ambiguous
- use 'helperText' to provide information on how to retrieve/look for that field

packages/forge/blocks/nocodb/actions/listTableRecords.ts
```diff
@@ -0,0 +1,105 @@
...
+    fields: option.string.layout({
+      label: 'Fields',
+      helperText:
+        'Comma-separated list of fields to include in the response. Leave empty to include all fields.',
+    }),
+    sort: option.string.layout({
+      label: 'Sort',
+      helperText:
+        'Comma-separated list of fields to sort the records by. Prefix with "-" to sort in descending order.',
+    }),
```
baptisteArno, line 25: I guess this could be a string option type (for better UI)

baptisteArno (changes requested):
That's awesome! Thank you so much. Just a few changes I'd do

baptisteArno (approved):

baptisteArno:
I reduced the actions to only "get", "create" and "update".  And tried to make them less "raw" and more user friendly. It also came with changes on how the Forge library works :)

Thank you for the initial work! It helped a lot!

PR #1410, opened 2024-04-03
Docs update

Emmet-v15:
Since you proxy the prisma commands so they have the right env varaible for DATABASE_URL & the location for schema.prisma file


apps/docs/self-hosting/deploy/manual.mdx
```diff
@@ -43,7 +43,7 @@ cp .env.example .env
 
 <Note>
   The database user should have the `SUPERUSER` role. You can setup and migrate
-  the database with the `pnpm prisma generate && pnpm db:migrate` command.
+  the database with the `pnpm db:generate && pnpm db:migrate` command.
```
baptisteArno, line 46: In fact, I overlooked, the generate step can be removed as it will be automatically generate during build step.

```suggestion
  the database with the `pnpm db:migrate` command.
```

baptisteArno (changes requested):

Emmet-v15 (commented):
In that case we also don't need to add the extra package.json line

apps/docs/self-hosting/deploy/manual.mdx
baptisteArno, line 46: This still needs to be removed :)

```suggestion
  the database with the `pnpm db:migrate` command.
```

baptisteArno (changes requested):

PR #1413, opened 2024-04-04
Mention how not to overwrite GSheet results

tgdn:
The GSheet documentation didn't mention how not to overwrite existing rows. By indicating adding a unique ID as the first column, we make sure no rows are overwritten.


baptisteArno:
Insert row should not replace any existing row.

If you think you found a bug, open an issue 🙏

We also discuss it here: https://discord.com/channels/1155799591220953138/1225378078742544395/1225378078742544395

PR #1438, opened 2024-04-10
:children_crossing: Enable streaming for self-hosters

baptisteArno:


baptisteArno:
Cosing as the builder preview URL is not valid

PR #1478, opened 2024-04-26
Fix typo in troubleshooting

eliottha:
Hi Arno,

Thank you for your amazing project.

I have made the following small change in the docs:
- "signed" text should be "sign", fixed this small oversight in the docs.


baptisteArno:
Thank you! 👍

PR #1490, opened 2024-05-02
:zap: (variables) Add session option in variables

### LINKED ISSUE #1488: Add "session" option on set variable
This would be enabled by default to avoid storing in results unnecessary variable values.

Will be used by #1484. Because storing Transcript in results would not make sense.

baptisteArno:
Closes #1488


baptisteArno:
@coderabbitai review

Issue #584, opened 2023-06-30
href btn - open in same tab


antonymarion:
Hi Arnaud,

When we use the href feature here 

![image](https://github.com/baptisteArno/typebot.io/assets/8767496/d64058bf-df95-441e-96e0-49d19380763f)

Then it opens a new tab, on each link we have created.

Could it be possible to open the link in the same tab than the one hosting the whole app with its typebot?

Regards,
Antony

baptisteArno:
That would mean the user would leave the bot when clicking on a link, I don't understand why it would be something you'd want?

antonymarion:
Ah, very simple.

Imagine this bot helps to navigate inside a 3D scene.

Once the user has used the bot, he then can discover (thanks to the link) the places that fits his answers.

Thus he does not need the bot any more..

smitgol:
i will pick up this issue

baptisteArno:
Typebot is now in maintenance mode, so I’m no longer planning new features, product improvements or large internal projects. I’m closing this to keep the issue tracker focused on bugs, security and compatibility. Thanks for the suggestion!

PR #1534, opened 2024-05-23
Open href links in same tab

ankitts:
#584


baptisteArno:
This should be an option, not a default behavior

PR #1675, opened 2024-08-01
⚡️ Reset remembered state if the typebot is updated

### LINKED ISSUE #1427: Add an option to update typebots in existing sessions

### LINKED ISSUE #1508: Option to reset remembered state if the typebot was updated

basuabhirup:
Wipe stored chat state if a typebot is updated.

Resolves #1508 
Resolves #1427  

https://github.com/user-attachments/assets/7668d6a7-2916-4158-b715-1cccaeaf6ddf


Dimpipay (approved):

packages/embeds/js/src/components/Bot.tsx
```diff
@@ -150,8 +150,24 @@ export const Bot = (props: BotProps & { class?: string }) => {
       const initialChatInStorage = getInitialChatReplyFromStorage(
         data.typebot.id
       )
-      if (initialChatInStorage) {
-        setInitialChatReply(initialChatInStorage)
+      if (
+        initialChatInStorage &&
+        initialChatInStorage.sessionUpdatedAt &&
```
baptisteArno, line 155: I guess we don't need to use a `sessionUpdatedAt` field. just comparing the `typebot.updatedAt` fields should be enough (API response vs in storage), if they are different then we restart the chat

baptisteArno (changes requested):

packages/embeds/js/src/components/Bot.tsx
basuabhirup: Yeah right, that would make more sense 💯 

Actually, I started to implement the same logic initially in the backend part but then realized the importance of accessing the `initialChatInStorage` object and hence moved the same logic to the frontend, that's why I compared it with the `sessionUpdatedAt` field.

Implemented the changes!

baptisteArno:
Added documentation and bumped embed libs version. Thank you for your contrib once again @basuabhirup you rock 🔥

baptisteArno (approved):

PR #1732, opened 2024-08-21
Update configuration.mdx

CptPlastic:
Added help link for OAuth consent screen when people are setting up google sheets access they are forgetting to add the scopes to the screen itself.


baptisteArno:
Cheers!

PR #1772, opened 2024-09-05
feat: add switch to disable button validation

gabscar:
Adicionado um swith para remover a validação de retentiva do botão


baptisteArno:
Any valid use case where this would be needed?

gabscar:
> Any valid use case where this would be needed?

I made an enhancement that allows the user to customize the validation message for the buttons. The option to disable this feature can be useful in cases where manual validation of the variable sent by the user is desired.

For example, if the user types a text in WhatsApp because they don't want any of the options, I might try to route that user to an agent.

Issue #1219, opened 2024-02-05
Option to add small buttons below the text input


baptisteArno:
It would serve as suggestions for the user to select

![Image](https://github.com/user-attachments/assets/4f77d505-5fad-49d3-b823-d3fe546c348e)

baptisteArno:
Typebot is now in maintenance mode, so I’m no longer planning new features, product improvements or large internal projects. I’m closing this to keep the issue tracker focused on bugs, security and compatibility. Thanks for the suggestion!

PR #1775, opened 2024-09-06
Feat/customize retry button message

gabscar:
I made an enhancement that allows the user to customize the validation message for the buttons. The option to disable this feature can be useful in cases where manual validation of the variable sent by the user is desired.

For example, if the user types a text in WhatsApp because they don't want any of the options, I might try to route that user to an agent.


baptisteArno:
I appreciate this.

I understand it makes sense for WhatsApp. But on the web this option does not make sense as we can't type anything we want.

What I envisioned is adding buttons with the Text input #1219 and this would match the WhatsApp behavior.

So I think Buttons input should always be strict in its validation. And if you need to enable free text from user you'd use instead the Text input with that suggestion buttons enabled.

Do you see what I mean?

gabscar:
Understanding that this feature would not directly affect the web, as the user will never have the input to type available.

This option only benchmarks external systems to give the user more freedom both to remove the options validation loop and to customize the validation message.

baptisteArno:
This is crucial to have consistent behaviors on all runtimes. People not using WA at all might not understand that issue. Or even if they understand it would just add useless infor in the block settings, impacting their experience. 🙏

I appreciate the work done though and we can totally adapt this for https://github.com/baptisteArno/typebot.io/issues/1219

aaag1980:
Good morning. I can't use Typebot and migrate my bots and start using it precisely because there is no "no match" path. It is essential that there is an exit option for when the user does not want any button options. For example, I have a bot in both MS Teams and on the web, where I offer buttons. However, the user may want to type something instead of selecting the buttons, where I do the appropriate treatment (optional buttons). I currently use Voiceflow and I can't migrate to Typebot because of this functionality. 
Screenshot_20241206_085534_WhatsApp
Screenshot_20241206_085517_WhatsApp

PR #1778, opened 2024-09-08
FEATURE: Zendesk chat

jwalsh-vori:
This adds a Zendesk Chat block to allow chat flows to open a Zendesk live chat window, similar to the Chatwoot block.


packages/forge/blocks/zendeskChat/actions/authenticate.ts
```diff
@@ -0,0 +1,50 @@
...
+      if (!email || email.length === 0
+        || !userId || userId.length === 0
+        || !name || name.length === 0
+        || secretKey === undefined
+        || keyId === undefined
+        || tokenVariableId === undefined
+      ) return
+
+      var token = jwt.sign({ scope: 'user', external_id: userId, name: name, email: email, email_verified: emailVerified }, secretKey, { algorithm: "HS256", keyid: keyId });
+      variables.set(tokenVariableId, token);
```
baptisteArno, line 46:
```suggestion
      variables.set(tokenVariableId, jwt.sign({ scope: 'user', external_id: userId, name: name, email: email, email_verified: emailVerified }, secretKey, { algorithm: "HS256", keyid: keyId }));
```

packages/forge/blocks/zendeskChat/index.ts
```diff
@@ -0,0 +1,14 @@
+import { createBlock } from '@typebot.io/forge'
+import { ZendeskChatLogo } from './logo'
+import { auth } from './auth'
+import { openMessenger } from './actions/openMessenger'
+import { authenticate } from './actions/authenticate'
+
+export const zendeskChatBlock = createBlock({
+  id: 'zendesk-chat',
+  name: 'Zendesk Chat',
```
baptisteArno, line 9: Can we rename it to just "Zendesk" I might add more actions in the future that are not tied to Chat

packages/forge/blocks/zendeskChat/logo.tsx
```diff
@@ -0,0 +1,37 @@
...
+    height="32px"
+    viewBox="0 0 256.00 256.00"
+    version="1.1"
+    xmlns="http://www.w3.org/2000/svg"
+    xmlnsXlink="http://www.w3.org/1999/xlink"
+    preserveAspectRatio="xMidYMid"
+    fill="#000000"
+    stroke="#000000"
+    stroke-width="0.00256"
+  >
```
baptisteArno, line 15:
```suggestion
  <svg
    viewBox="0 0 256.00 256.00"
    {...props}
  >
```

packages/forge/blocks/zendeskChat/logo.tsx
```diff
@@ -0,0 +1,37 @@
...
+    <g id="SVGRepo_bgCarrier" stroke-width="0" />
+
+    <g
+      id="SVGRepo_tracerCarrier"
+      stroke-linecap="round"
+      stroke-linejoin="round"
+    />
+
+    <g id="SVGRepo_iconCarrier">
+      {' '}
```
baptisteArno, line 25: Remove all `{' '}` if possible 🙏

packages/forge/blocks/zendeskChat/actions/authenticate.ts
```diff
@@ -0,0 +1,50 @@
...
+      isRequired: true,
+    }),
+    email: option.string.layout({
+      label: 'Email',
+      isRequired: true,
+    }),
+    emailVerified: option.string.layout({
+      label: 'Email Is Verified',
+      defaultValue: "false"
+    }),
```
baptisteArno, line 24:
```suggestion
    wasEmailVerified: option.boolean.layout({
      label: 'Email was verified',
    }),
```

I think it is clearer

packages/forge/blocks/zendeskChat/actions/authenticate.ts
```diff
@@ -0,0 +1,50 @@
+import { createAction, option } from '@typebot.io/forge'
+import jwt from 'jsonwebtoken';
```
baptisteArno, line 2:
```suggestion
import { sign } from 'jsonwebtoken';
```

packages/forge/blocks/zendeskChat/actions/openMessenger.ts
```diff
@@ -0,0 +1,65 @@
+import { createAction, option } from '@typebot.io/forge'
+import { auth } from '../auth'
+
+export const openMessenger = createAction({
+  auth,
+  name: 'Open Messenger',
```
baptisteArno, line 6: Is Messenger the right word. Is it what Zendesk using?

packages/forge/blocks/zendeskChat/actions/openMessenger.ts
```diff
@@ -0,0 +1,65 @@
...
+  options: option.object({
+    key: option.string.layout({
+      label: "Web Widget Key",
```
baptisteArno, line 9: Can you avoid capital case for option fields name? This would be better if it is "Web widget key" instead.

packages/forge/blocks/zendeskChat/actions/openMessenger.ts
```diff
@@ -0,0 +1,65 @@
...
+      isRequired: true,
+    }),
+    enableAuth: option.boolean.layout({
```
baptisteArno, line 12:
```suggestion
    isAuthEnabled: option.boolean.layout({
```

packages/forge/blocks/zendeskChat/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "license": "AGPL-3.0-or-later",
+  "devDependencies": {
+    "@typebot.io/forge": "workspace:*",
+    "@typebot.io/lib": "workspace:*",
+    "@typebot.io/tsconfig": "workspace:*",
+    "@types/react": "18.2.15",
+    "typescript": "5.4.5"
+  },
+  "dependencies": {
+    "@types/jsonwebtoken": "9.0.2",
```
baptisteArno, line 16: This should be in devDependencies

packages/forge/blocks/zendeskChat/index.ts
baptisteArno: Also change id to zendesk. Make sure to search and replace in the entire project 🙏

baptisteArno (changes requested):
That's awesome, thank you so much!!

Can you also provide a quick demo video to this PR? You rock

baptisteArno:
@coderabbitai review

packages/forge/blocks/zendeskChat/index.ts
jwalsh-vori: I originally wanted to do exactly that, however, there are two separate API families for Zendesk, the core Zendesk API, https://vorihealth.zendesk.com/admin/apps-integrations/apis/zendesk-api/settings, which covers their Tickets, and other CRM functions and and their Conversations API -  https://vorihealth.zendesk.com/admin/apps-integrations/apis/conversations-api, which is the Sunshine conversations product they acquired. They are really two separate apps, and don't share API keys, etc. So, now sure if it would be confusing to do this and manage credentials as one single integration. What do you think?

packages/forge/blocks/zendeskChat/actions/openMessenger.ts
jwalsh-vori: Zendesk is a bit confusing here. Messaging is a Channel for them and their Web Widget is one of the UIs available for Messaging, along with others like FB Messenger, WhatsApp, etc.

Screenshot 2024-09-10 at 10 29 53 PM

Screenshot 2024-09-10 at 10 27 24 PM


So, this could be changed to openWebWidget. We could also add additional actions for the other messaging channels in the future.

packages/forge/blocks/zendeskChat/actions/authenticate.ts
jwalsh-vori: Email Verification is a status of the email, so it's probably better to describe it in terms of the current state. How about 'isEmailVerified' and keep the label the same?

packages/forge/blocks/zendeskChat/actions/openMessenger.ts
jwalsh-vori: renamed action to openWebWidget

jwalsh-vori:
> That's awesome, thank you so much!!
> 
> Can you also provide a quick demo video to this PR? You rock

https://www.loom.com/share/2696180db3f84367ae56bed5dc8959f2

packages/forge/blocks/zendeskChat/index.ts
jwalsh-vori: I updated it to Zendesk Messaging to be more aligned to the product. If there is a way to manage multiple credentials without confusing the user, happy to move to just Zendesk. Thoughts?

baptisteArno: I honestly think it makes more sense to name it Zendesk and update the credentials form accordingly, explaining that signing key is only useful for authenticating users before opening up the Zendesk Chat widget and that it is optional.

In the future we will add the fields for the other Zendesk APIs all of which will be optional as well.

Does that make sense?

packages/forge/blocks/zendeskChat/actions/authenticate.ts
baptisteArno: Ok, but still convert it to boolean and non-title case?

packages/forge/blocks/zendeskChat/index.ts
jwalsh-vori: Makes sense to me. What I'm going to do is close this PR so I can generate the block properly as Zendesk. I'll add a new PR for that. Thanks!

packages/forge/blocks/zendeskChat/actions/authenticate.ts
jwalsh-vori: @baptisteArno a question on the boolean. If I change it to boolean, it doesn't appear that I can set that value from a variable, which I would need to be able to do as we would be tracking as part of a user profile. Am I missing something? Otherwise, I think we need to leave it as a string so we can do that.

jwalsh-vori:
Close in favor of https://github.com/baptisteArno/typebot.io/pull/1794

PR #1784, opened 2024-09-11
#1769 Fixed: Http request on Search Bar

SyedAbuBakerAli:
Fixed: #1769 Issue now we getting http request when we type HTTP on SearchBar

Screenshot 2024-09-11 211222

Note: currently we use http with webshook name in code If we changed WEBHOOK constant to HTTP_REQUEST then on search when we type HTTP then we got on search without modifing search so if you want I will do it.


baptisteArno:
This should not be hard-coded. We should instead from BlockLabel.tsx create a mapping between block type and name and use the name for the search

baptisteArno:
I deployed the fix yesterday!

PR #1786, opened 2024-09-11
(variables) - regex change for variables prefixed with dollars

### LINKED ISSUE #1782: Dollar prefix with a variable in text bubble does not appear
https://discord.com/channels/1155799591220953138/1281842592056868894/1281842592056868894

ahsmha:
Fixes #1782

### DIFF
--- a/packages/variables/parseVariables.ts
+++ b/packages/variables/parseVariables.ts
@@ -23,6 +23,9 @@ const inlineCodeRegex = /\{\{=(.+?)=\}\}/g
 // {{variable}} and ${{{variable}}}
 const variableRegex = /\{\{([^{}]+)\}\}|(\$)\{\{([^{}]+)\}\}/g
 
+// {{variable}}
+const textVariableRegex = /\{\{([^{}]+)\}\}/g
+
 export const parseVariables =
   (
     variables: Variable[],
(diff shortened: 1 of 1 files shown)


baptisteArno:
Not good you ignore the case `${{{Variable}}}`

ahsmha:
@baptisteArno what could be the appropriate fix over here?
and when is ${{{var}}} case used?

ahsmha:
One workaround is that we could store dollar sign inside a variable and then use it alongside the value without space in between. But that is just a workaround

packages/variables/parseVariables.ts
```diff
@@ -23,6 +23,9 @@ const inlineCodeRegex = /\{\{=(.+?)=\}\}/g
 // {{variable}} and ${{{variable}}}
 const variableRegex = /\{\{([^{}]+)\}\}|(\$)\{\{([^{}]+)\}\}/g
```
baptisteArno, line 24: You basically ignored that regex. We need to extend it instead of replacing it with a simpler one

CptPlastic:
I look at the `$` as currency so maybe just use a bunch of them and that is the variable 

you could set a list. and only accept that in your regex. I fear that in some instances the $ will break things and would be best practice to wrap currency in its own variable 
```
Kč
₡
RD$
$
```

packages/variables/parseVariables.ts
ahsmha: ```\{\{([^{}]+)\}\}|(\$)\{\{\{([^{}]+)\}\}\}```
@baptisteArno what do you think of this one?
so for input - 
```
${{xyz}} 
${{{abc}}} 
{{pqr}}
```

it gives following matches - 
```
{{xyz}}
${{{abc}}}
{{pqr}}
```

i wonder if it breaks anything?

baptisteArno: I fixed the issue already 🙏

https://github.com/baptisteArno/typebot.io/commit/3c07041f2f0a03852138b953a51838b81b077895

PR #1791, opened 2024-09-13
Fix editable overflowing in the variables drawer

tgdn:
In the `VariablesDrawer`, long variables overflow their parent container by pushing away action buttons as can be seen in the following screenshot: 

image

Since we are in a flex context, the default value for `min-width` is `auto` which causes the container to grow automatically based on its content.

To make sure the `text-overflow` / `overflow` pair work correctly, and display an ellipsis, the parent container must have a fixed size. This can be solved by setting the `min-width` of the overflowing container to `0` (the default outside a flex context).

After applying `min-width: 0` to the editable component: 

image


baptisteArno:
Thanks! :)

PR #1794, opened 2024-09-14
FEAT_ Zendesk

jwalsh-vori:
Adds a new Zendesk block with initial support for launching the Zendesk Messaging Web Widget


jwalsh-vori:
@baptisteArno replacement PR for Zendesk

apps/docs/editor/blocks/integrations/zendesk.mdx
```diff
@@ -0,0 +1,43 @@
...
+    src="/images/blocks/integrations/zendesk/zendesk-end-user-auth.png"
+    alt="End User Authentication"
+  />
+</Frame>
+
+Learn more here: https://developer.zendesk.com/documentation/zendesk-web-widget-sdks/sdks/web/enabling_auth_visitors/#generating-a-signing-key
+
+## Authenticate Messaging User
+
+This action will generate a JWT token that can be passed as a variable to Open Web Widget you to authenticate a user in Zendesk. It requires the `User ID`, `Name` `Email`. Optionally, you can also set `Email Is Verified` if the email address has been verified. This is recommended to be set if appropriate.
```
baptisteArno, line 26: I found https://developer.zendesk.com/documentation/zendesk-web-widget-sdks/sdks/web/enabling_auth_visitors/#email-verification

> Zendesk will only look for verified email addresses in a JWT. If the email_verified claim is not set to true, the email address in the JWT will be ignored, and it will not be included in the end user's profile in Agent Workspace.

So I guess the Email Is Verified field is useless here. We just need to tell the user to make sure the email was verified priori to providing it to Typebot. We could put that in a moreInfoTooltip for the email field

Does that make sense?

baptisteArno (changes requested):
Ok sorry about this but I think the Authenticate User action is useless, it should be directly embedded in the Open Web Widget action. It would definitely make more sense.

apps/docs/editor/blocks/integrations/zendesk.mdx
jwalsh-vori: I originally thought the same, however, because open web widget is a client side action, that would lead to exposing the secrets used to create the JWT token on the client which would be highly insecure, no?

Email Verified is used in the JWT signing process. It's possible and valid to use verified and unverified emails when launching the web widget however as you may be dealing with new customers as well as existing, verified customers so we need to support verified and unverified emails. What verification does is that it tells the agent on the other end of the chat that the customer is known.

jwalsh-vori: @baptisteArno what are your thoughts on this?

baptisteArno: > I originally thought the same, however, because open web widget is a client side action, that would lead to exposing the secrets used to create the JWT token on the client which would be highly insecure, no?

You should be able for that action to define a server run function **AND** and a web function. The server function will first be executed.

> It's possible and valid to use verified and unverified emails when launching the web widget

But Zendesk is saying that if we provide Email without the email_verified field set to true then it will completely ignore the Email. So my question is what is the point to provide the email_verified field on Typebot. Let's make it always true and document the fact that the creator of the bot verifies the user email before providing it to Zendesk block

Or am I missing something?

jwalsh-vori: > You should be able for that action to define a server run function **AND** and a web function. The server function will first be executed.
> 

Oh, I see so, would the run block for openWebWidget look like this then? How do I pass the token from the server side auth function to the client openWebWidget function as it doesn't appear that variables are available to the client functions?

`run: {
    server: async ({
      credentials: { conversationsSecretKey, conversationsKeyId },
      options: { userId, name, email, isEmailVerified, tokenVariableId },
      variables,
    }) => {
      if (!email || email.length === 0
        || !userId || userId.length === 0
        || !name || name.length === 0
        || conversationsSecretKey === undefined
        || conversationsKeyId === undefined
        || tokenVariableId === undefined
      ) return
      variables.set(tokenVariableId, sign({ scope: 'user', external_id: userId, name: name, email: email, email_verified: "true" }, conversationsSecretKey, { algorithm: "HS256", keyid: conversationsKeyId }));
    },
    web: {
      parseFunction: ({ options }) => {
        return {
          args: {
            isAuthEnabled: options.isAuthEnabled ? options.isAuthEnabled?.toString() : "false",
            token: options.token ?? null,
            key: options.key ?? null
          },
          content: parseOpenMessenger()
        }
      }
    },
  },`

baptisteArno: Right, we would still need the user to provide a variable for where to store the token. Then on the server you set the variable and you use it in the client.

baptisteArno: In fact parseFunction is executed on the server so you are right it should have access to variables, credentials etc... And you won't even have to declare a server run function. Let me push an update, will let you know once you can merge main into this branch 🙏

baptisteArno: Done: https://github.com/baptisteArno/typebot.io/commit/f613ce366bf048eda9c4d29b37a5f4327814be3e

jwalsh-vori: NICE!! Thank you!

jwalsh-vori: Ok, have removed the auth action, updated the openWebWidget action to generate the token. Looks good!

jwalsh-vori: and email_verified is set to true now

baptisteArno (changes requested):
Awesome, that's much better.

I think we can definitely simplify this even more! 😅

1. Make userId, name and email optional
2. Remove isAuthEnabled ( because auth is implicitely enabled as soon as userId is provided)

baptisteArno (approved):

baptisteArno:
Amazing work. Thank you @jwalsh-vori 🔥

jwalsh-vori:
> Amazing work. Thank you @jwalsh-vori 🔥

You are very welcome, thank you for all the great suggestiosn. Could we get this added to a new release of self-hosted?

baptisteArno:
This will be added to next monthly release 1st of October. But I can make an exception if you really need this now now?

jwalsh-vori:
> This will be added to next monthly release 1st of October. But I can make an exception if you really need this now now?

No, as long as it's available for self-hosters in Oct, that's perfect, thank you!

Issue #1792, opened 2024-09-13
The "create new variable" button does not appear when results match name of new variable


tgdn:
**Describe the bug**
The "create new variable" button does not appear when results match name of new variable.

**To Reproduce**
1. Create a new Typebot
2. Open the variables drawer
3. Create a new variable "maet"
4. Attempt to create a new variable "ae"
5. Since the name of the variable "ae" matches "maet", it is impossible to create a new variable with the name "ae"

This is an issue because affiliate parameters may not be updated, and thus the name of the variable may not be chosen in advance.

⚠️ It is actually possible to press enter and create the variable, but it is IMHO not very user-friendly. It would make more sense to display the "+" button to click to create the variable.

**Expected behavior**
Render the "+" button to create a new variable whenever the user has typed something in the variable search input.

PR #1795, opened 2024-09-14
fix: #1792

steninkgeorge:
fixed the problem labeled in #1792.


baptisteArno:
Thank you but that is not it. We should not show the button if the search input match exactly one of the existing variables name

PR #1803, opened 2024-09-24
Update s3.mdx

ricardolucena:
Update CORS json format https://docs.aws.amazon.com/AmazonS3/latest/userguide/ManageCorsUsing.html


baptisteArno:
It's basically the same, when configuring on AWS you have to give the content of `CORSRules`. But on some other services you need to provide the entire CORS policy that needs to include CORSRules.

PR #1895, opened 2024-11-22
🐛 Fix issue - Phone country prefix input changing once one is already set

### LINKED ISSUE #1894: Phone country prefix is not changing once one is already set

basuabhirup:
Fix #1894

https://github.com/user-attachments/assets/60403582-7141-4be7-b8bb-faa6b4ec3b02


baptisteArno:
Thanks @basuabhirup. I just refactored it for better clarity IMHO 🙏

PR #1945, opened 2025-01-04
Update google-sheets.mdx

jimit24365:
While selfhosting we need to ensure to enable Google Picker API in-oder to avoid issue of Invalid API key error Reference : https://stackoverflow.com/questions/22435410/google-drive-picker-developer-key-is-invalid-error

This is solution to the problem I mentioned over Discord https://discord.com/channels/1155799591220953138/1157296381908746281/1323022912369852459


baptisteArno:
This is not appropriate since this comment is towards self-hosters only. This is the doc about the Google Sheets blocks for any users.

Issue #1957, opened 2025-01-16
Condition block content overflow issue


baptisteArno:
![CleanShot 2025-01-16 at 09 12 46@2x](https://github.com/user-attachments/assets/e6f434a6-998e-44bd-a66d-870f98572946)

PR #1959, opened 2025-01-16
Fixed: Condition block content overflow issue

Nil2000:
Fixes: #1957 

Before fix
image

After fix
image

**Also fixed the extra width taken by the container**

Hope this helps

### DIFF
--- a/apps/builder/src/features/blocks/logic/condition/components/ConditionContent.tsx
+++ b/apps/builder/src/features/blocks/logic/condition/components/ConditionContent.tsx
@@ -54,7 +54,6 @@ export const ConditionContent = ({
                 px={1.5}
                 py={0.5}
                 fontWeight={"medium"}
-                boxShadow={"md"}
               >
                 {variable.name}
               </Box>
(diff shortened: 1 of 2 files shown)


apps/builder/src/features/blocks/logic/condition/components/ConditionContent.tsx
```diff
@@ -38,9 +45,18 @@ export const ConditionContent = ({
...
+                bgColor="orange.400"
+                color="white"
+                borderRadius="0.375rem"
+                px={1.5}
+                py={0.5}
+                fontWeight={"medium"}
+              >
                 {variable.name}
-              </Tag>
+              </Box>
```
baptisteArno, line 59: This cannot be done with Tag component?

baptisteArno (changes requested):

apps/builder/src/features/blocks/logic/condition/components/ConditionContent.tsx
Nil2000: Nope text was clipped.

Nil2000: image

baptisteArno:
Managed to do it with word-break prop: 7f29c5822cc4e627bb7f8895d34720b825f2c413

PR #1960, opened 2025-01-17
FEAT_POSTHOG

jwalsh-bdt:
Add Posthog integration block to provide posthog features:

- Identify
- Identify Group
- Capture
- PageView
- Get Flag

updated docs to include Posthog documentation

updated posthog node sdk version in telemetry


packages/forge/blocks/posthog/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "exports": {
+    ".": "./src/index.ts",
+    "./schemas": "./src/schemas.ts"
+  },
+  "devDependencies": {
+    "@typebot.io/forge": "workspace:*",
+    "@typebot.io/tsconfig": "workspace:*"
+  },
+  "dependencies": {
+    "posthog-node": "^4.3.2"
```
baptisteArno, line 15:
```suggestion
    "posthog-node": "4.3.2"
```

We try to pin to specific versions on all the deps in Typebot :)

packages/forge/blocks/posthog/src/logo.tsx
```diff
@@ -0,0 +1,53 @@
+/** @jsxImportSource react */
+
+export const PosthogLogo = (props: React.SVGProps<SVGSVGElement>) => (
+  <svg width="25px" height="15px" viewBox="0 0 25 15" version="1.1">
```
baptisteArno, line 4:
```suggestion
  <svg viewBox="0 0 25 15" {...props}>
```

This only need viewbox and spread the props to be displayed properly

packages/telemetry/package.json
```diff
@@ -11,7 +11,7 @@
     "@typebot.io/prisma": "workspace:*",
     "@typebot.io/zod": "workspace:*",
     "ky": "1.2.4",
-    "posthog-node": "3.1.1"
+    "posthog-node": "^4.3.2"
```
baptisteArno, line 14:
```suggestion
    "posthog-node": "4.3.2"
```

packages/forge/blocks/posthog/src/createClient.ts
```diff
@@ -0,0 +1,5 @@
+import { PostHog } from "posthog-node";
```
baptisteArno, line 1: Can we put both `createClient` and `createProperties` in a `helpers` folder?

baptisteArno (changes requested):
That's great work 🙌 Thank you.

We always try to map the API directly with Typebot. And a Page view is ultimately calling the capture function. So can we remove the Page view action and maybe add a dropdown on the capture event where we can choose between "Page view" or "Custom". Heck we could even fetch the existing events and display them dynamically 🤯

packages/forge/blocks/posthog/package.json
jwalsh-bdt: Updated

packages/forge/blocks/posthog/src/logo.tsx
jwalsh-bdt: Done

packages/telemetry/package.json
jwalsh-bdt: Done

packages/forge/blocks/posthog/src/createClient.ts
jwalsh-bdt: Done

jwalsh-bdt:
> That's great work 🙌 Thank you.
> 
> We always try to map the API directly with Typebot. And a Page view is ultimately calling the capture function. So can we remove the Page view action and maybe add a dropdown on the capture event where we can choose between "Page view" or "Custom". Heck we could even fetch the existing events and display them dynamically 🤯

Good call. I added a Custom event and Page view into Capture and also added Screen view and Survey Sent. Those seem to be the main options that Posthog offers for now. I wasn't able to figure out how to only show the 'Event Name' option when Custom is selected, any suggestions there? I did try using a discriminatingUnion but that makes it very clumsy in the UI and only seems to support an array vs a single option being selected. I'm fine with it as is but wanted to check to see if I'm missing something.

baptisteArno:
I've updated the auth UX a bit and updated the logo so that it's better aligned.

I think it will be easier to exactly match the posthog-node API naming.

So we don't need to have a dropdown with custom, page view etc. We just need a text input called "Event". And also the other names should match

jwalsh-bdt (commented):
Done

jwalsh-bdt:
@baptisteArno - should be good to go, thanks for the suggestions and help!

baptisteArno:
@jwalsh-bdt I simplified and improved a bit Capture and IdentifyGroup.

I removed `identify` and `getFlag`.

- Identify on the server does not make much sense. Posthog says in their docs:
    ```
    Although it is technically possible to call identify using our backend SDKs, it is typically most useful in frontends. This is 
    because there is no concept of anonymous sessions in the backend SDKs.
    ```
    So I added `personProperties` prop in capture action to be able to set person properties alongside an event.

- getFlag unfortunately can't be implemented like this because you are saving the JSON payload in a variable which is something we don't promote on Typebot. So we'd need to be able to Test the feature flag to be able to map payload items to individual variables. Which is something we can't do for now with Forged block. But in the future we'll be able to do that :)

jwalsh-bdt:
@baptisteArno on the getFlag item, it seemed to work fine when the JSON is a simple string, I was able to use it to return different variants of a message to be displayed, would be great to support this as I think this is actually a very powerful use case for a tool like typebot where you want to test different messaging to improve conversion or control branching.

baptisteArno:
Ok indeed that would be nice. I will check it out https://github.com/baptisteArno/typebot.io/issues/1980

PR #2000, opened 2025-02-07
Fixed: Default phone country resetted when emptying the input

### LINKED ISSUE #1999: Default phone country resetted when emptying the input

ompathak2004:
Fixed #1999  Prevent selected country code from resetting to INT when input is cleared

- Removed the block that reset **selectedCountryCode** to **INT** when the input field was cleared.

- This ensures that if a country is selected, clearing the input does not revert it to **INT**.


baptisteArno:
Cheers!

baptisteArno (approved):

PR #2004, opened 2025-02-10
Updated docker-compose

sebiweise:
Just made some changes to the docker-compose file.


docker-compose.yml
```diff
@@ -17,24 +25,38 @@ services:
...
-      typebot-db:
-        condition: service_healthy
-    restart: always
     ports:
       - '8081:3000'
-    env_file: .env
+
+networks:
+  typebot-network:
+    driver: bridge
```
baptisteArno, line 58: why?

docker-compose.yml
```diff
@@ -17,24 +25,38 @@ services:
...
+    healthcheck:
+      test: ["CMD-SHELL", "redis-cli ping | grep PONG"]
+      start_period: 20s
+      interval: 30s
+      retries: 5
+      timeout: 3s
+    volumes:
+      - redis-data:/data
+    networks:
+      - typebot-network
```
baptisteArno, line 42: Redis is **very** optional for Typebot for now so I'm not sure it makes sense to add it here

docker-compose.yml
```diff
@@ -17,24 +25,38 @@ services:
...
+
   typebot-builder:
+    <<: *typebot-common
```
baptisteArno, line 45: What are you trying to achieve with this typebot-common refactoring? What was not working beofre?

baptisteArno (changes requested):

docker-compose.yml
sebiweise: Oh okey I didn´t know that, I just implemented it and thought it would be nice to have in a Full-Stack-Example.

docker-compose.yml
sebiweise: This "typebot-common" is used to reduce the duplicated code that is needed to start the Builder and View container, this setup is working, is copied from my production version.
The only thing that was different between the Builder and the View was the image definition itself.

docker-compose.yml
baptisteArno: If I am correct, this is the default behavior so we don't really have to explicitely add this on each service

docker-compose.yml
```diff
@@ -17,24 +25,38 @@ services:
...
-      typebot-db:
-        condition: service_healthy
     ports:
       - '8080:3000'
-    extra_hosts:
-      - 'host.docker.internal:host-gateway'
-    env_file: .env
 
   typebot-viewer:
+    <<: *typebot-common
```
baptisteArno, line 51: missing

```yml
    extra_hosts:
      - 'host.docker.internal:host-gateway'
```

docker-compose.yml
baptisteArno: Ok why not!

baptisteArno (changes requested):

docker-compose.yml
sebiweise: I´m not sure what this is for?

docker-compose.yml
sebiweise: Yeah, that's right, it's just a preference of mine. If you want I can remove it for the Typebot repo.

docker-compose.yml
baptisteArno: Indeed, not useful anymore!

baptisteArno:
Thank you for this 🙏

Issue #2018, opened 2025-02-13
Be able to click on an image to zoom on it


baptisteArno:
Would be nice if as a user of the bot, I can click on an image bubble to open it in fullscreen

baptisteArno:
Yes but I want to use https://ark-ui.com/react/docs/components/dialog

PR #2035, opened 2025-02-22
Added full screen dialog for image bubble

Nil2000:
Partially Fixes: #2018 

[demo-full-screen_edit_0.webm](https://github.com/user-attachments/assets/68304949-b8bd-4d89-a013-d3ee0911063e)

Few issues I faced
1. I don't know why keeping `button` and also `img` inside `div` worked.
2. If zoomed after opening dialog, sometimes can be closed sometimes not.(Don't know why)


baptisteArno:
Indeed modal behavior was totally broken. I just fixed it! Thanks for the work

PR #2038, opened 2025-02-25
Updating

fazin-ahamed:


baptisteArno:
Careful with license compliance! https://docs.typebot.io/self-hosting/get-started#license-requirements

Issue #2013, opened 2025-02-10
Save language preference in database


baptisteArno:

bst1n:
thanks !

PR #2043, opened 2025-02-26
[Feat] Save language preference in database

enzogms:
Closes [2013](https://github.com/baptisteArno/typebot.io/issues/2013)


apps/builder/src/features/user/components/UserPreferencesForm.tsx
```diff
@@ -49,7 +49,7 @@ export const UserPreferencesForm = () => {
   };
 
   const updateLocale = (locale: keyof typeof localeHumanReadable) => () => {
-    document.cookie = `NEXT_LOCALE=${locale}; path=/; max-age=31536000`;
+    updateUser({ preferredLanguage: locale });
```
baptisteArno, line 52:
```suggestion
    document.cookie = `NEXT_LOCALE=${locale}; path=/; max-age=31536000`;
    updateUser({ preferredLanguage: locale });
```

Better to keep the cookie mechanism to make sure all the internal are working as expected with Next.js

packages/prisma/postgresql/schema.prisma
```diff
@@ -60,6 +60,7 @@ model User {
   displayedInAppNotifications Json?
   credentials                 UserCredentials[]
   groupTitlesAutoGeneration   Json?
+  preferredLanguage           String?                   @default("en")
```
baptisteArno, line 63:
```suggestion
  preferredLanguage           String?
```

No need for a default value in DB

baptisteArno (changes requested):
I think it would be best to 100% rely on the NEXT_LOCALE cookie.

I would prefer having a redirection done in `UserProvider` if current language is not the preferred one.

Also make sure to add `preferredLanguage` in clientUserSchema to have a correct type on client-side

enzogms:
@baptisteArno done ✅

PR #2048, opened 2025-03-02
fix: Docker configuration

mathieu-coingt:
Missing options like 'minio' dependency for 'caddy-gen' and 'restart: always' on 'minio' services.

### DIFF
--- a/apps/docs/self-hosting/deploy/docker.mdx
+++ b/apps/docs/self-hosting/deploy/docker.mdx
@@ -192,6 +192,7 @@ If you don't already have an S3 storage available, you could include it in your 
 services:
   minio:
     image: minio/minio
+    restart: always
     command: server /data
     ports:
       - '9000:9000'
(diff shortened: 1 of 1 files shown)


apps/docs/self-hosting/deploy/docker.mdx
```diff
@@ -203,6 +204,7 @@ services:
   # This service just makes sure a bucket with the right policies is created
   createbuckets:
     image: minio/mc
+    restart: always
```
baptisteArno, line 207: Not necessary as it is supposed to run once and quit.

baptisteArno (changes requested):

apps/docs/self-hosting/deploy/docker.mdx
baptisteArno: So it needs to be removed :)

baptisteArno:
Feel free to reopen

PR #2078, opened 2025-03-17
chore(blocks, number): migrate NumberInput to ArkUI

alexis-falaise:
Related to #2067


baptisteArno (changes requested):
I see that you copy/pasted the same class names as the number input from landing page.

Do you think it's possible to export NumberInput into `packages/ui/components` and re-use it on both apps?

Maybe not since it also accepts variable string. So maybe at least reuse the class names?

baptisteArno:
Will do it another way, splitting it into 2 separate components: `NumberInput` and `NumberInputWithVariableButton`

PR #2080, opened 2025-03-17
feat(blocks, number): add format options inputs for number blocks

alexis-falaise:
Following the display capabilities of Intl.NumberFormat, this allows user to define their number blocks formatting options.

The options are then used to parse number into the expected string format for bubbles.

This adds a **Format** section featuring:
- A style choice between `decimal`, `percentage`, `currency` and `unit`
- A currency dropdown with all supported currencies
- A unit dropdown with all supported units
- A locale input to override the default locale

Capture d’écran 2025-03-17 à 17 57 29

Capture d’écran 2025-03-17 à 17 57 21


packages/embeds/js/src/features/blocks/inputs/number/components/NumberInput.tsx
```diff
@@ -14,24 +16,22 @@ type NumberInputProps = {
...
-  const checkIfInputIsValid = () =>
-    inputRef?.value !== "" && inputRef?.reportValidity();
-
   const submit = () => {
-    if (checkIfInputIsValid())
+    if (!numberInput().invalid) {
       props.onSubmit({
         type: "text",
-        value: inputRef?.value ?? inputValue().toString(),
+        value: numberInput().valueAsNumber.toFixed(DEFAULT_PRECISION),
```
baptisteArno, line 32: Why not:

```suggestion
        value: numberInput().valueAsNumber.toString(),
```

So that we don't have to provide a precision?

packages/embeds/js/src/features/blocks/inputs/number/components/NumberInput.tsx
```diff
@@ -50,36 +50,53 @@ export const NumberInput = (props: NumberInputProps) => {
...
-          onInput={(e) => {
-            setInputValue(targetValue(e.currentTarget));
-          }}
-          type="number"
-          min={props.block.options?.min}
-          max={props.block.options?.max}
-          step={props.block.options?.step ?? "any"}
+          min={minValue}
+          max={maxValue}
+          step={stepValue}
```
baptisteArno, line 89: I think this needs to be provided to `useNumberInput` in order to work

packages/embeds/js/src/features/blocks/inputs/number/components/NumberInput.tsx
```diff
@@ -14,24 +16,22 @@ type NumberInputProps = {
...
+    },
+  });
   let inputRef: HTMLInputElement | undefined;
 
-  const checkIfInputIsValid = () =>
-    inputRef?.value !== "" && inputRef?.reportValidity();
-
   const submit = () => {
-    if (checkIfInputIsValid())
+    if (!numberInput().invalid) {
```
baptisteArno, line 29: We used `reportValidity` to have at least the Browser system message displayed like this:

CleanShot 2025-03-18 at 16 06 09@2x

Is there a way to still have that even when using thie Ark component?

baptisteArno (changes requested):
Awesome work. Thank you so much! Just a few improvements could be made. Open to discuss :)

apps/builder/src/features/blocks/inputs/number/components/NumberInputSettings.tsx
```diff
@@ -66,6 +122,78 @@ export const NumberInputSettings = ({ options, onOptionsChange }: Props) => {
...
+        <AccordionItem>
+          <AccordionButton>
+            <Text>
+              {t("blocks.inputs.number.settings.format.label")}
+              {` (${formatAccordionLabel})`}
+            </Text>
+          </AccordionButton>
+          <AccordionPanel>
+            <DropdownList
+              label={t("blocks.inputs.number.settings.style.label")}
```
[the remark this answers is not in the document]
alexis-falaise: I tend to make things super explicit 😄 
In this case it works well right below the "Format" accordion title indeed

apps/builder/src/features/blocks/inputs/number/components/NumberInputSettings.tsx
```diff
@@ -14,13 +37,28 @@ type Props = {
...
     onOptionsChange({
       ...options,
       labels: { ...options?.labels, placeholder },
     });
   const handleButtonLabelChange = (button: string) =>
     onOptionsChange({ ...options, labels: { ...options?.labels, button } });
+  const handleCurrencyChange = (currency: string) =>
+    onOptionsChange({
+      ...options,
+      currency: currency ?? defaultNumberInputOptions.currency,
```
baptisteArno, line 60:
```suggestion
      currency: currency
```

We never save the default options, we keep it undefined and apply it at runtime

apps/builder/src/features/blocks/inputs/number/components/NumberInputSettings.tsx
```diff
@@ -30,6 +68,25 @@ export const NumberInputSettings = ({ options, onOptionsChange }: Props) => {
...
     step?: NonNullable<NumberInputBlock["options"]>["step"],
   ) => onOptionsChange({ ...options, step });
+  const handleStyleChange = (style: NumberInputStyle) =>
+    onOptionsChange({
+      ...options,
+      style,
+      currency:
+        style === NumberInputStyle.CURRENCY
+          ? defaultNumberInputOptions.currency
+          : undefined,
```
baptisteArno, line 78: Maybe best:

```suggestion
      currency:
        style === NumberInputStyle.CURRENCY
          ? options.currency
          : undefined,
```

packages/embeds/js/src/features/blocks/inputs/number/components/NumberInput.tsx
```diff
@@ -14,24 +18,46 @@ type NumberInputProps = {
...
+      ? Number(props.block.options.min)
+      : undefined;
+  const maxValue =
+    props.block.options?.max !== undefined
+      ? Number(props.block.options.max)
+      : undefined;
+  const stepValue =
+    props.block.options?.step !== undefined
+      ? Number(props.block.options.step)
+      : undefined;
```
baptisteArno, line 32: We tend to only declare variables if it's re-used at least twice. Otherwise, we can inline all of them

packages/blocks/inputs/src/number/currencies.ts
```diff
@@ -0,0 +1,320 @@
+export enum Currency {
```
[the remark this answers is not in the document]
baptisteArno: I meant, we should avoid having 2 different files that have the same data, so I suggest we import the existing payment/currencies file here for the Number Input settings

apps/builder/src/i18n/en.json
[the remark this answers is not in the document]
baptisteArno: We really need just the en.json. The other translations will be done on Tolgee

packages/embeds/js/src/features/blocks/inputs/number/components/NumberInput.tsx
baptisteArno: Still, it doesn't work because ArkUI under the hood is not providing the native HTML attributes to the number input. Can we think of another way to display the error message? For now maybe with a error toast? You can use:

```tsx
toaster.create({
              description,
            })
```

baptisteArno (changes requested):

apps/builder/src/features/blocks/inputs/number/components/NumberInputSettings.tsx
alexis-falaise: In the case of the currency setting, leaving it undefined can crash the app

apps/builder/src/features/blocks/inputs/number/components/NumberInputSettings.tsx
```diff
@@ -14,13 +37,28 @@ type Props = {
...
   const { t } = useTranslate();
+
+  useEffect(() => {
+    if (!options?.locale) {
+      const browserLocale = navigator.language;
+      if (browserLocale.match(localeRegex)) {
+        onOptionsChange({ ...options, locale: browserLocale });
+      }
+    }
+  }, []);
```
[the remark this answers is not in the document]
alexis-falaise: This should run only once, resolving

apps/builder/src/features/blocks/inputs/number/components/NumberInputSettings.tsx
baptisteArno: That's why we do it at runtime:

```tsx
const numberInput = useNumberInput({
    formatOptions: {
      style: props.block.options?.style,
      currency: props.block.options?.currency ?? defaultNumberInputOptions.currency,
      unit: props.block.options?.unit,
    },
    min: minValue,
    max: maxValue,
    step: stepValue,
  });
```

baptisteArno (approved):

tgdn:
@baptisteArno Ever since today all our inputs that use the number block are invalid numbers that include a comma.

baptisteArno:
> @baptisteArno Ever since today all our inputs that use the number block are invalid numbers that include a comma.
> 
> @alexis-falaise

Checking

baptisteArno:
@tgdn I can't reproduce. Can you provide a reproduction?

tgdn:
1. Create a number input 
image

2. Enter a number and display it:
image


Whenever the number is submitted to an API, the number is considered invalid since it's an invalid number

tgdn:
Before, numbers submitted to an API were valid numbers: 
image


Now they are submitted as a formatted number: 
image


@baptisteArno

baptisteArno:
Ok make sense. Will push a fix asap

baptisteArno:
Should be fixed, can you try again @tgdn?

baptisteArno:
Ok I pushed another fix just now

tgdn:
@baptisteArno both bugs were resolved 🙌 thank you!

PR #2084, opened 2025-03-19
feat(system-messages): add offline error system messages

alexis-falaise:
This adds new system messages that will be shown in a toast in case of a network failure.

| Builder | Viewer |
|--------|--------|
| Capture d’écran 2025-03-19 à 16 17 31| Capture d’écran 2025-03-19 à 16 17 58|


packages/settings/src/constants.ts
```diff
@@ -30,6 +30,8 @@ export const defaultSettings = {
 export const defaultSystemMessages = {
   invalidMessage: "Invalid message. Please, try again.",
   botClosed: "This bot is now closed",
+  offlineErrorTitle: "You are offline",
```
baptisteArno, line 33: I'd prefer "Network error" or something more generic as the error checking does not include only when the user is offline

packages/settings/src/schemas.ts
```diff
@@ -8,6 +8,8 @@ import { rememberUserStorages } from "./constants";
 export const systemMessagesSchema = z.object({
   invalidMessage: z.string().optional(),
   botClosed: z.string().optional(),
+  offlineErrorTitle: z.string().optional(),
+  offlineErrorMessage: z.string().optional(),
```
baptisteArno, line 12:
```suggestion
  networkErrorTitle: z.string().optional(),
  networkErrorMessage: z.string().optional(),
```

To be more generic :)

packages/embeds/js/src/utils/error.ts
```diff
@@ -0,0 +1,22 @@
...
+      "NetworkError",
+      "Network request failed",
+      "Connection refused",
+      "ECONNREFUSED",
+      "timeout exceeded",
+    ];
+
+    return networkErrorMessages.some((message) =>
+      error.message.includes(message),
+    );
```
baptisteArno, line 14: Where does this list come from?

Should we lowercase the messages when checking or are we sure about the casing of these messages?

baptisteArno (changes requested):

packages/settings/src/constants.ts
alexis-falaise: Agreed 👌

packages/embeds/js/src/utils/error.ts
alexis-falaise: The list is not comprehensive as I couldn't find a definitive reference on the matter. Casing should only help though, updated

PR #2098, opened 2025-03-25
Use value for Button & Picture choice blocks

### LINKED ISSUE #1713: Separate Variable Values from Display Text on Button or Picture Choice blocks
When using Buttons or Picture Choice Blocks, the display text is what is stored as the variable value. This makes variable values unwieldy, especially if the display text is long. Also, if a flow is being copied to be localized in another language, any logic that evaluates the variable value has to also be updated to use the localized value. 

While it is possible to set the value using a set variable block on each choice, that adds complexity to the flow. It would be idea if we could have an ability to set the variable value for each input choice, so that the display text is what the users selects but the input value is what is stored in a variable. This would ideally be a constant value, so instead of using "Knee Pain" as the value to drive logic, it would use KNEE_PAIN.

See attached images for what this might look like. In the first image, there is a button choice input where a patient chooses which issue they are suffering from. In the 2nd image, I show that for Knee Pain, the health_issue variable value is set to KNEE_PAIN. In the third image, I show how the condition logic evaluates KNEE_PAIN, not "Knee Pain". 

This way, when I make a copy of the flow to localize it to another language, only the display text has to change, not the logic.

This could be done as an optional setting perhaps, where the setting of variable value defaults to the display text, unless a value is set at the choice level so that existing flows are not impacted

image

image

image

alexis-falaise:
This adds a value field for Button and Picture choice blocks. If a value is defined, it will take precedence over the button content or the image title for variable evaluation.

Capture d’écran 2025-03-25 à 15 14 20
Capture d’écran 2025-03-25 à 15 13 47

Closes #1713


apps/builder/src/features/blocks/inputs/buttons/components/ButtonsItemSettings.tsx
```diff
@@ -53,6 +60,14 @@ export const ButtonsItemSettings = ({ item, onSettingsChange }: Props) => {
           onConditionChange={updateDisplayCondition}
         />
       </SwitchWithRelatedSettings>
+      <TextInput
+        label={t("blocks.inputs.button.buttonSettings.buttonValue.label")}
+        helperText={t(
+          "blocks.inputs.button.buttonSettings.buttonValue.helperText",
+        )}
```
baptisteArno, line 67: Usually we prefer to use `moreInfoTooltip` when we want to explain what the input is about. `helperText` is for when we need to provide indication on where to find the value for that input

apps/builder/src/i18n/en.json
```diff
@@ -121,6 +121,8 @@
   "blocks.events.start.eventCard.tooltip": "Already added in the bot flow.",
   "blocks.inputs.button.addItem.ariaLabel": "Add item",
   "blocks.inputs.button.buttonSettings.addComparisonButton.label": "Add comparison",
+  "blocks.inputs.button.buttonSettings.buttonValue.label": "Button value:",
```
baptisteArno, line 124: I'm wondering, maybe "Internal value" would make more sense? What do you think

packages/embeds/js/src/components/InputChatBlock.tsx
```diff
@@ -93,10 +93,11 @@ export const InputChatBlock = (props: Props) => {
...
     )?.formattedMessage;
+
     if (formattedMessage && props.block.type !== InputBlockType.FILE)
       setAnswer((answer) =>
         answer?.type === "text"
-          ? { ...answer, label: formattedMessage }
+          ? { ...answer, label: answer.label ?? formattedMessage }
           : answer,
       );
   });
```
baptisteArno, line 103: This can indeed be generalized, if the input has a label applied, we shouldn't overwrite it

```suggestion
    if (answer.label) return
    const formattedMessage = formattedMessages().findLast(
      (message) => props.chunkIndex === message.inputIndex,
    )?.formattedMessage;

    if (formattedMessage)
      setAnswer((answer) =>
        answer?.type === "text"
          ? { ...answer, label: answer.label ?? formattedMessage }
          : answer,
      );
  });
```

packages/embeds/js/src/components/InputChatBlock.tsx
```diff
@@ -168,7 +169,9 @@ const Input = (props: {
   onSubmit: (answer: InputSubmitContent) => void;
   onSkip: (label: string) => void;
 }) => {
-  const onSubmit = (answer: InputSubmitContent) => props.onSubmit(answer);
+  const onSubmit = (answer: InputSubmitContent) => {
+    props.onSubmit(answer);
+  };
```
baptisteArno, line 174: Indeed it seems this func is not useful. Let's remove it :)

packages/embeds/js/src/features/blocks/inputs/buttons/components/Buttons.tsx
```diff
@@ -26,11 +26,16 @@ export const Buttons = (props: Props) => {
...
+    const item = filteredItems()[itemIndex];
+    const { value, content } = item;
+
     props.onSubmit({
       type: "text",
-      value: filteredItems()[itemIndex]?.content ?? "",
+      value: value ?? content ?? "",
+      label: content ?? value ?? "",
     });
+  };
```
baptisteArno, line 38: The issue is that we now always provide a label even if value is not defined by the user so the formatting from server will never be applied.

I think that'd be best to only add a label if the input has indeed defined internal values

packages/bot-engine/src/blocks/inputs/buttons/parseButtonsReply.ts
```diff
@@ -20,76 +91,8 @@ export const parseButtonsReply = (
...
-  if (!matchedItem) return { status: "fail" };
-  return {
-    status: "success",
-    content: matchedItem.content ?? "",
-    outgoingEdgeId: matchedItem.outgoingEdgeId,
-  };
+
+  return block.options?.isMultipleChoice
+    ? handleMultipleChoice(displayedItems, inputValue)
+    : handleSingleChoice(displayedItems, inputValue);
```
baptisteArno, line 97: Even clearer?

```suggestion
    ? parseMultipleChoiceReply(displayedItems, inputValue)
    : parseSingleChoiceReply(displayedItems, inputValue);
```

packages/bot-engine/src/blocks/inputs/buttons/parseButtonsReply.ts
```diff
@@ -1,9 +1,80 @@
...
+import {
+  getItemContent,
+  matchByIndex,
+  matchByKey,
+  sortByContentLength,
+} from "../../../helpers/choiceMatchers";
 import type { ParsedReply } from "../../../types";
 import { injectVariableValuesInButtonsInputBlock } from "./injectVariableValuesInButtonsInputBlock";
 
+const createMatchedItemsResponse = (
```
baptisteArno, line 13: I'd inline that function, IMHO that's not a necessary abstraction here

packages/embeds/js/src/features/blocks/inputs/buttons/components/Buttons.tsx
alexis-falaise: Yes that makes sense here.

How should we ideally deal with the MultiplePictureChoice case ?

pnpm-lock.yaml
```diff
@@ -0,0 +1,472 @@
+lockfileVersion: '6.0'
```
baptisteArno, line 1: You ran pnpm install by mistake 😄

packages/embeds/js/src/features/blocks/inputs/buttons/components/MultipleChoicesForm.tsx
```diff
@@ -42,17 +42,24 @@ export const MultipleChoicesForm = (props: Props) => {
...
+      value: selectedItems
+        .map((item) => {
+          return item?.value ?? item?.content ?? "";
+        })
+        .join(", "),
+      label: selectedItems
+        .map((item) => {
+          return item?.content ?? item?.value ?? "";
+        })
         .join(", "),
```
baptisteArno, line 60: I'd say, check if one of the selected items has an internal value, if that's the case, we set the label attribute otherwise we set it to undefined. Same for multiple picture choices

baptisteArno (changes requested):

baptisteArno (approved):

PR #2100, opened 2025-03-26
[Feat] Add "Disable public URL" option

### LINKED ISSUE #1240: Add "Disable public URL" option

enzogms:


baptisteArno (changes requested):
I appreciate the attempt but unfortunately what you implemented will simply make the bot not startable anywhere if this option is enabled. To make it work we need to check the API request origin and check if it matches.

Also it made me think this is not a good option "Disable public URL". Instead, it would be better to just let the user add the allowed public URL origin like typebot.co or typebot.io

PR #2113, opened 2025-04-01
Fix: Ensure color picker displays above other elements

### LINKED ISSUE #2110: Color picker overlap issue
Image

seraph14:
Closes #2110 

## Issue
When an accordion is toggled open, the color picker popup was overlapping with other UI elements (specifically Roundness and Thickness controls).

## Solution
This PR ensures the color picker appears above all other elements by:
- Using Portal to render the popover content outside the normal DOM hierarchy

## Implementation Notes
- The original popover placement was "right" is kept and now it is displayed on the right.
- This approach preserves the existing functionality while fixing the overlay issue.
- No additional dependencies were required for this fix.

## Screenshots

image


baptisteArno (changes requested):
https://github.com/user-attachments/assets/61ec127f-18c8-4ee9-a203-af2582c61183

Closes its parent popover if any

PR #2114, opened 2025-04-01
Reply event block

### LINKED ISSUE #1592: Add "On new message" event block
Would allow the user to kill the session if a particular command is sent on WhatsApp for example.

alexis-falaise:
This adds a new type of event, triggered on user replies.

An exit condition can be defined to decide if the end of the alternative flow should exit the bot.

Capture d’écran 2025-04-08 à 11 01 36

Closes #1592


baptisteArno (changes requested):
Testing on the Lead generation template: [typebot-export-lead-generation-rnlagdn.json](https://github.com/user-attachments/files/19647648/typebot-export-lead-generation-rnlagdn.json), it doesn't seem to work as expected!

https://github.com/user-attachments/assets/b7f7193f-e571-4cb1-9063-05c51b4abe2f

packages/bot-engine/src/continueBotFlow.ts
```diff
@@ -97,6 +101,14 @@ export const continueBotFlow = async (
     });
   }
 
+  if (reply?.type === "text") {
```
baptisteArno, line 104: Why not "audio" as well? The only reply type that should not trigger this event IMO would be command

alexis-falaise: Sure, we can add all but command 👌

alexis-falaise:
> Testing on the Lead generation template: [typebot-export-lead-generation-rnlagdn.json](https://github.com/user-attachments/files/19647648/typebot-export-lead-generation-rnlagdn.json), it doesn't seem to work as expected!
> 
>  CleanShot.2025-04-08.at.12.51.56.mp4

I couldn't reproduce that issue exactly, but I realized that we were jumping a block in the portal - which could be related.
Let me know if the last fix works better !

packages/events/src/schemas.ts
```diff
@@ -25,17 +27,39 @@ export const commandEventSchema = eventBaseSchema.extend({
...
+  commandEventSchema,
+  replyEventSchema,
+]);
+
 export type TEvent = z.infer<typeof eventSchema>;
 
 export type TEventWithOptions = Extract<TEvent, { options?: any }>;
 
-export const draggableEventSchema = draggableEventSchemas[0];
 export type TDraggableEvent = z.infer<typeof draggableEventSchema>;
```
baptisteArno, line 65:
```suggestion
const draggableEventSchemas = [commandEventSchema, replyEventSchema] as const;

export const eventSchema = z.discriminatedUnion("type", [
  startEventSchema,
  ...draggableEventSchemas,
]);
export type TEvent = z.infer<typeof eventSchema>;

export type TEventWithOptions = Extract<TEvent, { options?: any }>;

export const draggableEventSchema = z.discriminatedUnion("type", [
  ...draggableEventSchemas
]);
export type TDraggableEvent = z.infer<typeof draggableEventSchema>;
```

baptisteArno: I like to do this formatting:

```ts
const schema = ...
type Schema = z.infer<>
```

Sticked together

packages/typebot/src/schemas/edge.ts
```diff
@@ -23,5 +24,6 @@ export const edgeSchema = z.object({
   id: z.string(),
   from: sourceSchema,
   to: targetSchema,
+  condition: conditionSchema.optional(),
```
baptisteArno, line 27: I think it makes more sense to add the condition to the sessionState schema instead as it does not really make sense in the global edgeSchema scope

As we discussed: queuedEdgeIds would then have the edgeId and a condition

packages/bot-engine/src/computeCurrentProgress.ts
```diff
@@ -93,7 +93,7 @@ const computePossibleNextInputBlocks = ({
   if (outgoingEdgeIds.length > 0 || group.blocks.length !== blockIndex + 1)
     return possibleNextInputBlocks;
 
-  if (typebotsQueue.length > 1 || typebotsQueue[0].queuedEdgeIds) {
+  if (typebotsQueue.length > 1 || typebotsQueue[0].queuedEdgeIds?.length) {
```
baptisteArno, line 96: why?

packages/bot-engine/src/events/executeResumeAfter.ts
```diff
@@ -0,0 +1,67 @@
...
+  event,
+  sessionStore,
+}: {
+  state: SessionState;
+  event: TEvent;
+  sessionStore: SessionStore;
+}) => {
+  if (!state.currentBlockId) return state;
+
+  const portalEdge =
```
baptisteArno, line 21: A more understandable name would be `nextBlock` maybe?

packages/bot-engine/src/events/executeResumeAfter.ts
```diff
@@ -0,0 +1,67 @@
+import type { SessionState } from "@typebot.io/chat-session/schemas";
+import type { Condition } from "@typebot.io/conditions/schemas";
+import { EventType } from "@typebot.io/events/constants";
+import type { TEvent } from "@typebot.io/events/schemas";
+import { getBlockById } from "@typebot.io/groups/helpers/getBlockById";
+import type { SessionStore } from "@typebot.io/runtime-session-store";
+import { addPortalEdge } from "../addPortalEdge";
+import { getNextBlockById } from "../getNextGroup";
+
+export const executeResumeAfter = async ({
```
baptisteArno, line 10: I think this naming is very abstract. Maybe `connectEdgeToNextBlock`?

packages/bot-engine/src/events/executeEvent.ts
```diff
@@ -0,0 +1,41 @@
...
+import type { SessionStore } from "@typebot.io/runtime-session-store";
+import { addBlockToTypebotIfMissing } from "../addBlockToTypebotIfMissing";
+import { getNextGroup } from "../getNextGroup";
+type Props = {
+  state: SessionState;
+  event: TDraggableEvent;
+  sessionStore: SessionStore;
+};
+
+export const executeEvent = async ({ state, event, sessionStore }: Props) => {
```
baptisteArno, line 13: Maybe a more descriptive name would be `updateCurrentBlockIdWithEvent`

packages/bot-engine/src/updateVariablesInSession.ts
```diff
@@ -5,12 +5,107 @@ import type {
...
+            ? foundVariable.value.concat(reply.text)
+            : reply.text,
+      },
+    ],
+    currentBlockId: undefined,
+    state,
+  });
+
+  return updatedState;
+};
```
baptisteArno, line 107: Since all these helpers only called once, I think it's preferable to inline them even if it seems a bit verbose

baptisteArno (changes requested):

packages/bot-engine/src/getNextGroup.ts
```diff
@@ -141,6 +166,52 @@ export const getNextGroup = async ({
   };
 };
 
+export async function getNextBlockById(
```
baptisteArno, line 169: Is it possible to move this to executeResumeAfter file? Since it's only used there for now it can be a local function

packages/bot-engine/src/computeCurrentProgress.ts
alexis-falaise: The empty `queudEdgeIds` was leading to a non-verbose 500

packages/typebot/src/schemas/edge.ts
alexis-falaise: An array of ids is a good practice for a store, I would advocate for keeping it as is:
- better for lookup performances
- respects the semantics

From my understanding an edge is situational, and there seems to be a strong 1-1 relationship between an edge and a condition to execute it. Do you foresee cons to this approach that I could have missed ?

packages/bot-engine/src/events/executeResumeAfter.ts
alexis-falaise: By the function name yes, but the object contains a block itself so I thought that would be confusing 😅

packages/bot-engine/src/updateVariablesInSession.ts
alexis-falaise: Each of them are used twice (in the `updateInputVariablesInSession` and in `saveVariables` file)

packages/bot-engine/src/computeCurrentProgress.ts
baptisteArno: queuedEdgeIds should never be empty though, that's why I'm confused. Can you still reproduce the 500 bug?

packages/typebot/src/schemas/edge.ts
baptisteArno: It's all about context, with your current implementation, a user could see in the API reference that a typebot edge can have a condition, he would ask himself what it is about, why an edge can have a condition. And then, if he adds a condition to any edge (through the API), should it then always be checked when we trigger that edge? That's not the current implementation though.

Whereas if we see that queuedEdges contains the edgeId and a condition object we can easily guess that the condition is triggered once we are popping out the queuedEdge. The only place where this condition matter is in this particular process

We never process, search, filter `queuedEdgeIds` so I am not sure about performance? Am I missing something?

packages/bot-engine/src/computeCurrentProgress.ts
baptisteArno: Oh no my bad it can actually be empty!

packages/bot-engine/src/events/executeResumeAfter.ts
baptisteArno: Right, then why not `resumeMetadata` 🚀

packages/typebot/src/schemas/edge.ts
alexis-falaise: Let's avoid making it publicly available then, I've made the updates. ✅

baptisteArno (approved):

baptisteArno (approved):

PR #2115, opened 2025-04-02
Update route.ts to include error handling and use edge to improve response

danmainah:
Edge Runtime Utilization - This utilizes edge to reduce start times and improve response latency.

Error Handling: Use try catch to ensure response in case of an error and also standardize responses using createErrorResponse


baptisteArno:
Please open an issue first to talk about the problem you wan to solve. Here you are updating an old API route.

The new one is: https://github.com/baptisteArno/typebot.io/blob/d9df90c9b1434a285347223a23e857bf1671a9fc/apps/viewer/src/app/api/v2/sessions/%5BsessionId%5D/streamMessage/route.ts

PR #2122, opened 2025-04-09
Command event exit condition

alexis-falaise:
This replaces "Resume after" with "Exit condition" in command event block, leveraging the new conditional queued edge mecanism.

command


baptisteArno:
Closing in favor of the new [Return block](https://docs.typebot.io/editor/blocks/logic/return#return)

PR #2192, opened 2025-05-30
fix((setVariable): server-side template code injection

odaysec:
https://github.com/baptisteArno/typebot.io/blob/492e365fc543e5664d36cefd9e9b3bbf54072e0d/packages/deprecated/typebot-js/src/iframe/index.ts#L52-L52

Fix the issue need to eliminate the use of `Function(data.codeToExecute)` and replace it with a safer alternative. The best approach is to validate and restrict the input to a predefined set of safe operations or commands. This ensures that only known, trusted code can be executed. 

Steps to implement the fix:
1. Replace the dynamic evaluation of `data.codeToExecute` with a mechanism that maps predefined commands to specific functions or actions.
2. Introduce a whitelist of allowed commands and validate `data.codeToExecute` against this whitelist.
3. If `data.codeToExecute` does not match any allowed command, ignore it or log an error.

Directly evaluating user input (an HTTP request parameter) as code without properly sanitizing the input first allows an attacker arbitrary code execution. This can occur when user input is treated as JavaScript, or passed to a framework which interprets it as an expression to be evaluated. Examples include AngularJS expressions or JQuery selectors.

### DIFF
--- a/packages/deprecated/typebot-js/src/iframe/index.ts
+++ b/packages/deprecated/typebot-js/src/iframe/index.ts
@@ -49,6 +49,17 @@ const processMessage = (
   if (data.redirectUrl) window.open(data.redirectUrl);
   if (data.newVariableValue && callbacks.onNewVariableValue)
     callbacks.onNewVariableValue(data.newVariableValue);
-  if (data.codeToExecute) Function(data.codeToExecute)();
+  if (data.codeToExecute) {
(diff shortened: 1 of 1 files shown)


baptisteArno:
This lib is deprecated and very outdated.

Make sure to create an issue before creating a PR 🙏

PR #2208, opened 2025-06-08
♻️ Clean up Jump block references when deleting edges

### LINKED ISSUE #2198: Jumps at deleted block keeps the block id and breaks the bot
**Describe the bug**
If you create a Jump to a group and provide a specific block, and then delete that block, the id remains. When the user reaches that block, the bot breaks and finishes the session.

I believe this is a bug because the same thing does not happen if you delete a group. The group's id is removed from the Jump.

**To Reproduce**
1. Create a new jump to a specific block

Image

2. Delete that block

Image

5. See that the deleted block's id is still there

Image

**Expected behavior**
For the deleted block's id to be removed from the Jump, or a warning to be shown when the bot is published.

defsanmith:
Closes #2198


defsanmith:
@baptisteArno 

The cleanup has been implemented only for the Jump block as of now. Please let me know if there are any other instances that need this as well.

baptisteArno:
I don't think it's a good idea to silentely clean it. See: https://github.com/baptisteArno/typebot.io/issues/2198#issuecomment-2944049661 🙏

PR #2235, opened 2025-06-29
Fix MinIO bucket creation script for newer `mc` versions

adspacheco:
Replace deprecated `mc config host add` with `mc alias set`.

## Context

Running the current **createbuckets** service logs errors similar to:

```
mc: <ERROR> `config` is not a recognized command. Get help using `--help` flag.
mc: <ERROR> Unable to set anonymous `public` for `minio/typebot/public`. Requested path `/minio/typebot/public` not found.
```

These arise because the command `mc config host` add was removed in recent releases in favor of `mc alias set`. Without a valid alias, subsequent commands reference a non‑existent path and fail.

`--ignore-existing` – makes bucket creation idempotent so the service can run multiple times without failing.

### DIFF
--- a/apps/docs/self-hosting/deploy/docker.mdx
+++ b/apps/docs/self-hosting/deploy/docker.mdx
@@ -194,7 +194,7 @@ services:
     image: minio/minio
     command: server /data
     ports:
-      - '9000:9000'
+      - "9000:9000"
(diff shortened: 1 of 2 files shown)


baptisteArno:
Thanks!

Issue #1225, opened 2024-02-07
User "idle" event


stgcompany:
I think would be important to be able to set a timeout on every type of input blocks.

Lets say a contact takes longer than 5 minutes to answer... We could set this to be the timeout, and then execute a different path on the flow, maybe even executing some kind of function (as I referenced on the issue #1224).

In my opinion that'd improve a lot the automated conversation development process.

vforvilela:
I'm also interested in a similar feature, albeit for a distinct purpose. It's a common practice for users to compose and send messages in fragments when using integrated platforms like WhatsApp. They often write a phrase, send it, and then repeat the process for the next segment.

In light of this, my concept involves creating a message buffer. The idea is to accumulate messages in the buffer while the user is actively typing. Once the user pauses for a few seconds, the system would then process and send the entire buffered message. 

This approach aims to enhance the user experience by minimizing interruptions and facilitating a more seamless messaging flow.

Something like this:
![image](https://github.com/baptisteArno/typebot.io/assets/80433106/8ffcb108-40d1-41bb-900e-f8b8818b95bc)

souzadeveloper:
Good morning, no solution yet for the above suggestions and questions?

vforvilela:
> did you find any solution for this?

I had to resign from TypeBot due to the lack of features and bug fixes, like this one.

gabrielthe:
Any ideia, when will this feature be ready? It's a real need. Looking forward to it. @baptisteArno 
Thanks

Shiba1805:
Any news on this?

Jorgeaceiro:
I'm also looking to move away from typebot due to the lack of solutions like this
a timeout in variable collect

heloineto:
I also want this feature. It would be very helpful for our operation

baptisteArno:
Typebot is now in maintenance mode, so I’m no longer planning new features, product improvements or large internal projects. I’m closing this to keep the issue tracker focused on bugs, security and compatibility. Thanks for the suggestion!

PR #2241, opened 2025-07-04
Add per-input timeout configuration to trigger in events timeout (#1225)

ceconelo:
As discussed in #1225, in real-world conversations, it is crucial to handle situations where the user stops interacting. 

image

image

image

Feature Avaliable on:
- Text
- Buttons


clairton:
@baptisteArno its awesome!

baptisteArno:
Sorry but I don't like this implementation, I find the UX weak!

Please make sure to first open an issue describing what you'd like to solve and we can discuss the specs there 🙏

maikejo:
@ceconelo The solution you created only works in Web mode. I performed a test in Web mode, but it works perfectly, but nothing happens when the flow is done directly from WhatsApp. Would it be possible for it to work in WhatsApp mode?

ceconelo:
> Sorry but I don't like this implementation, I find the UX weak!
> 
> Please make sure to first open an issue describing what you'd like to solve and we can discuss the specs there 🙏

All good?

There are already several issues that, in some way, point to this need — in fact, I mentioned one of them in this PR.

The problem to be solved:

Take a specific action if the user does not interact with an input.

Example:
You request some information from the client, and if they don’t reply within x time, the chat should proceed to another flow.

ceconelo:
> @ceconelo The solution you created only works in Web mode. I performed a test in Web mode, but it works perfectly, but nothing happens when the flow is done directly from WhatsApp. Would it be possible for it to work in WhatsApp mode?

All good?

Actually, I was just testing Typebot and, in the heat of the moment, created this feature out of necessity. But since we didn’t move forward with the tool, I didn’t continue developing it.

PR #2244, opened 2025-07-07
feature(dify): allow select blocking response_mode

CosmoAbdon:
When using `response_mode: 'streaming'`, we frequently encounter malformed JSON errors, especially with long responses or special characters that are not properly escaped. This leads to issues such as:

`Fail: While creating Dify chat message: Unterminated string in JSON at position 9582 (line 1 column 9583)`

### Example
image

This happens because Dify sends each node's data (such as `Workflow Started`, `Node Started`, `HTTP Request`, etc.) as streaming chunks, which mimics a verbose debug log.

However, this data can be incomplete or improperly escaped, causing JSON parsing failures when trying to construct a final message.

## Proposed Change

- Add an UI to select the response_mode

## Why This Matters

Using `response_mode: 'blocking'` avoids this issue entirely by only returning the final answer, which is already well-formed and easier to parse. However, to retain the streaming experience, these changes aim to improve fault tolerance without losing interactivity.

### New UI

image

####

Few questions I have on mind.

- When using blocking, shall we let the user allow more than 10s (default timeout) ?
- Shall we fall back to `response_mode: 'blocking'` when encountering malformed chunks ?
- Do we need to sanitize the streaming chunks before parsing ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
   "devDependencies": {
     "@biomejs/biome": "1.9.4",
     "@tolgee/cli": "2.8.2",
+    "@types/cookie": "^1.0.0",
     "husky": "9.1.7",
     "sherif": "1.4.0",
     "turbo": "2.5.3",


baptisteArno:
I'm sorry but we need to create an issue first!

I can see ai SDK has a Dify community provider: https://ai-sdk.dev/providers/community-providers/dify. We should use this!

PR #2247, opened 2025-07-09
fix(dify): switch to ai-sdk and prevent malformed JSON in streaming

CosmoAbdon:
## Context

When using `response_mode: 'streaming'` with Dify, we frequently encountered issues with **malformed JSON**, especially in longer responses or when special characters aren't properly escaped.

This typically results in parsing failures such as:

image

These issues occur because Dify streams every node's output (e.g., `Workflow Started`, `Node Started`, `HTTP Request`, etc.) as individual chunks — similar to debug logs. However, this streamed content is often partial or improperly escaped, leading to unstable JSON construction.

## What This PR Does

- Refactors the Dify block to utilize the standard AI-SDK helpers:
  - `runChatCompletionStream`
  - `runChatCompletion`
- Ensures **retro-compatibility** with existing block configurations.
  - If the block is already configured for streaming, it gracefully continues using it unless known issues arise.
- Improves error handling and response reliability when working with Dify in real-world workflows.

## Reference

This is an improved version of PR #2244, which was previously closed after review feedback.

> The updated implementation now uses the recommended SDK:  
> https://ai-sdk.dev/providers/community-providers/dify

## Checklist

- [x] Uses SDK helpers consistently (`runChatCompletion*`)
- [x] Fully backward-compatible
- [x] Tested with complex and nested workflows
- [x] Improved error resilience for malformed JSON in streaming

## Considerations

1. **Timeout Handling for Blocking Mode**  
   Since `response_mode: 'blocking'` may involve longer processing times (especially with complex workflows), should we expose a way for users to configure the timeout?  
   Currently, the timeout behavior is default at 10s.

2. **Partial Streaming Recovery (Future Idea)**  
   Instead of switching entirely to blocking mode in problematic cases, could we explore reconstructing valid JSON chunks from streaming data using a buffer and validator approach? This would preserve the advantages of streaming but with safer handling.

3. **Backward Compatibility**  
   The refactor maintains full retro-compatibility with existing workflows and configurations. But we may want to document somewhere (CHANGELOG?) that the internal provider logic has been migrated.


package.json
```diff
@@ -34,6 +34,7 @@
   "devDependencies": {
     "@biomejs/biome": "1.9.4",
     "@tolgee/cli": "2.8.2",
+    "@types/cookie": "^1.0.0",
```
baptisteArno, line 37: why?

packages/forge/blocks/difyAi/package.json
```diff
@@ -8,10 +8,11 @@
     "./schemas": "./src/schemas.ts"
   },
   "dependencies": {
-    "ky": "1.2.4",
+    "@ai-sdk/ui-utils": "1.2.2",
     "@typebot.io/forge": "workspace:*",
     "@typebot.io/lib": "workspace:*",
-    "@ai-sdk/ui-utils": "1.2.2"
+    "dify-ai-provider": "^0.1.6",
```
baptisteArno, line 14: Can you pin the version? 🙏

packages/forge/blocks/difyAi/src/actions/createChatMessage.ts
```diff
@@ -1,362 +1,261 @@
...
+          sessionStore,
+          isVisionEnabled: false,
+          messages: [
+            {
+              role: "user",
+              content: options.query,
+            },
+          ],
+          temperature: undefined,
+          tools: undefined,
```
baptisteArno, line 256: Can be omitted?

packages/forge/blocks/difyAi/src/actions/createChatMessage.ts
```diff
@@ -1,362 +1,261 @@
...
-  let jsonChunk = "";
+const options = option.object({
+  applicationId: option.string.layout({
+    isRequired: false,
+    label: "Application ID",
+    placeholder: "Fill the applicationId",
+    helperText: "Dify Application ID",
+    allowCustomValue: true,
+    defaultValue: "default-app-id",
+  }),
```
baptisteArno, line 140: I think application ID should be set on the Auth object, don't you think? I'm not sure

baptisteArno (changes requested):

packages/forge/blocks/difyAi/src/actions/createChatMessage.ts
CosmoAbdon: So, right now the runChatCompletionStream has this Props: 

https://github.com/baptisteArno/typebot.io/blob/96c08dc8b24354108b9534cc28a7cbbd7be99f49/packages/ai/src/runChatCompletionStream.ts#L15-L24

We could change the Props type to:

```ts
type Props = {
  model: LanguageModel;
  variables: VariableStore;
  messages: Message[];
  tools?: Tools | undefined;
  isVisionEnabled: boolean;
  temperature?: number | undefined;
  responseMapping: ChatCompletionOptions["responseMapping"] | undefined;
  sessionStore: SessionStore;
};
```

It would work the same way (since ? is undefined as default), and it would allow omitting those fields.
Shall I also do this?

packages/forge/blocks/difyAi/src/actions/createChatMessage.ts
CosmoAbdon: I pulled from the block straight to the Auth.
So we can use from the credentials OR stay the default value.

CosmoAbdon:
My only point that I didnt see clearly how I could make this happen..
In scenarios where the processing time exceeds the default threshold of 10 seconds, the request fails.

How we could improve that?

As example: 
image

I've did this test on the previous implementation, since we've been using the new runChatCompletion(Stream) I didnt.

Let me check if this issue is now solved or stay still.

baptisteArno:
I improved the retro compat logic by using zod transform function.

Also added conversation ID saving logic that was missing.

Removed also Application ID and User ID which are useless it seems. Now defaulting to `system@typebot.io` user name.

Let me know if it works for you

CosmoAbdon:
Hey! Got a bad week with tons of meetings/issues/warrooms.
I'll take a look and make tests today!

Sorry 4 the delay.

packages/ai/src/runChatCompletion.ts
```diff
@@ -13,9 +13,9 @@ type Props = {
   model: LanguageModel;
   variables: VariableStore;
   messages: Message[];
-  tools?: Tools | undefined;
+  tools: Tools | undefined;
   isVisionEnabled: boolean;
-  temperature?: number | undefined;
+  temperature: number | undefined;
```
CosmoAbdon, line 17: So we're going back to place in the object ({tools: undefined...}), aight?

packages/forge/blocks/difyAi/src/actions/createChatMessage.ts
```diff
@@ -167,85 +66,114 @@ export const createChatMessage = createAction({
...
-        responseMapping: normalizeResponseMappingForAI(options.responseMapping),
+        responseMapping: options.responseMapping,
         logs,
         sessionStore,
         tools: undefined,
         temperature: undefined,
         isVisionEnabled: false,
+        headers: {
+          "user-id": options.user ?? defaultUserId,
+          "chat-id": options.conversationVariableId
```
CosmoAbdon, line 98: F, I had fixed that when testing, but forgot to push when I did :(
Nice catch!

CosmoAbdon:
Worked as expected!

Created an simple chatbot using Dify (using master), after that switchted to this PR branch, no issues so far.

I've noted the mistakes I made (from file managment to apispec, next time I'll do it properly)

Thanks 4 u help.

packages/ai/src/runChatCompletion.ts
baptisteArno: Yes I prefer to let the consumer know that this prop should be explicitely defined

PR #2248, opened 2025-07-17
Fixing PHP deprecation warning

dev-cb:
In PHP 8.2+, creating properties on an object that are not declared in the class definition is deprecated. The fix is to explicitly declare the property in the class definition.


baptisteArno:
Appreciate it, thanks!

PR #2254, opened 2025-07-22
Fix: generate token without refresh by updating CreateTokenModal

### LINKED ISSUE #2239: Problem when trying to create more than one token without refreshing the page
**Describe the bug**
Problem when trying to create more than one token without refreshing the page

**To Reproduce**
1. Go to 'Settings & Members'
2. Click on 'Your Account'
3. Click on API tokens 'Create'
4. Type your unique name and 'Create token'
5. When creating the API token, try repeating the process without refreshing the page

Image

Rakshat28:
Problem:
After creating an API token, reopening the `CreateTokenModal` would still display the previously created token instead of the token creation form. This blocked users from generating multiple tokens in one session without refreshing the page.

Cause
The modal's internal state variables (`newTokenValue`, `name`, and `isSubmitting`) were not being reset when the modal was closed or reopened.

 Solution
Added a `useEffect` hook that resets the internal state when the modal is closed. This ensures the modal always shows the token creation form upon reopening, allowing users to create multiple tokens in one session.

Before: 
https://github.com/user-attachments/assets/76f18bdd-71d2-48c3-a598-45b809236f95

After:
https://github.com/user-attachments/assets/8dc2a0ef-0661-41ef-93d4-e521719a56fd

Fixes #2239

### DIFF
--- a/apps/builder/src/features/user/components/CreateTokenModal.tsx
+++ b/apps/builder/src/features/user/components/CreateTokenModal.tsx
@@ -16,7 +16,7 @@ import {
 } from "@chakra-ui/react";
 import { useTranslate } from "@tolgee/react";
 import type { FormEvent } from "react";
-import React, { useRef, useState } from "react";
+import React, { useRef, useState, useEffect } from "react";
 import { createApiTokenQuery } from "../queries/createApiTokenQuery";
 import type { ApiTokenFromServer } from "../types";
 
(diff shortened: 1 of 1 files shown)


Rakshat28:
Hi @baptisteArno ! Just wanted to kindly follow up on this PR. Let me know if any changes are needed, happy to help!

baptisteArno:
Thank you for the contribution but it is not right 🙏

We prefer to avoid useEffect as much as possible

PR #2311, opened 2025-11-26
fix: force Node.js 22 for Vercel deployment

adryserage:
## Summary
- Add `.nvmrc` file to force Vercel to use Node.js 22 instead of 24
- Fixes `isolated-vm` compilation failure due to C++20 requirement on Node.js 24

## Problem
Vercel now defaults to Node.js 24.x, but the `isolated-vm` package fails to compile because:
1. No prebuilt binaries exist for Node.js 24
2. The C++ compiler on Vercel's build environment doesn't support C++20 (required by Node.js 24's V8 headers)

Error: `#error "C++20 or later required."`

## Solution
Add `.nvmrc` with value `22` to ensure Vercel uses Node.js 22, which has prebuilt binaries for `isolated-vm`.

## Test plan
- [ ] Verify Vercel deployment succeeds with Node.js 22
- [ ] Confirm `isolated-vm` installs without compilation


baptisteArno:
In theory, having

```
"engines": {
    "node": "^22"
  },
```

in package.json should be enough 🤔

Maybe it should be added to apps/builder/package.json as well

baptisteArno:
Closing in favor of #2354. Thanks for the heads up :)

PR #2406, opened 2026-02-25
docs: warn about escaping $ in SMTP_PASSWORD

Teraformerrr:
## What does this PR do?

Adds documentation warning about the `$` character in `SMTP_PASSWORD` values.

## Problem

When `SMTP_PASSWORD` contains a `$` character, the dotenv CLI (used to start the app) passes values through shell interpolation. The `$` is treated as the start of a variable name, silently corrupting the password and causing `535 Incorrect authentication data` errors.

For example, `abc$defghij` silently becomes `abc` at runtime.

## Changes

- `.env.example`: Added SMTP configuration block with warning about `$` escaping
- `.env.dev.example`: Same SMTP block and warning added
- `apps/docs/self-hosting/configuration.mdx`: Added a `<Note>` with correct/incorrect examples after the SMTP table

## Related Issue

Fixes the issue reported by @alehostert regarding SMTP_PASSWORD dollar sign escaping.

### DIFF
--- a/.env.dev.example
+++ b/.env.dev.example
@@ -19,4 +19,14 @@ NODE_OPTIONS=--no-node-snapshot
 
 NEXT_PUBLIC_PARTYKIT_HOST=localhost:1999
 
+# SMTP Configuration (Email Auth & Notifications)
+# Note: If your SMTP_PASSWORD contains a $ character, escape it as \$ (e.g., abc&\$defghij)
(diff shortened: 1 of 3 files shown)


baptisteArno:
I appreciate the PR but I think it is out of scope to be honest, Typebot should not document how dotenv cli .env parsing works.

PR #2411, opened 2026-03-11
fix: release ExternalCopy handles to prevent native memory leak

harshp77:
## Summary
- `ExternalCopy` objects created via `isolated-vm` for object-type variables were never `.release()`d, leaking native memory (bytes-KB per variable per isolate invocation).
- Store the `ExternalCopy` reference and call `.release()` in a `finally` block so the backing store is freed even if `copyInto()` throws.
- Fixed in both `parseTransferrableValue` and `createHttpReqResponseMappingRunner` in `packages/variables/src/codeRunners.ts`.

## Test plan
- [ ] Verify existing code runner tests still pass
- [ ] Monitor memory usage under sustained isolate invocations to confirm leak is resolved


baptisteArno:
@codex review

baptisteArno:
cheers!

PR #2462, opened 2026-04-15
🐛 Fix SSRF safe dispatcher DNS lookup handling

baptisteArno:
- Fix `validatingLookup` to handle `{ all: true }` DNS lookup mode that undici passes, which returns an array of addresses instead of a single string
- Add localhost bypass in development mode to match existing `validateHttpReqUrl` behavior
- Without this fix, `fetch()` in Set Variable code blocks silently failed for external URLs


baptisteArno:
@codex review

PR #2463, opened 2026-04-16
fix: prevent abuse of preview chat endpoints via rate limiting

### LINKED ISSUE #2423: Using all of TypeBot's features in the preview.
**Describe the bug**
I noticed that when creating a /preview session in Typebot, we can use the infrastructure freely without needing to use the free plan.

**To reproduce**

1. When generating https://typebot.io/api/v1/typebots/<typebot ID>/preview/startChat
2. Then we can use https://typebot.io/api/v1/sessions/<session ID>/continueChat for free.

**Expected behavior**
Since it's just a preview, shouldn't there be a usage limitation?

utsav1213:
Fixes #2423

### 🐛 Problem

Preview chat endpoints (`/preview/startChat` and `/sessions/continueChat`) bypass usage tracking and do not increment `totalChatsUsed`.

This allows users to abuse preview sessions as production chats without hitting plan limits.

---

### ✅ Solution

Introduced a Redis-backed rate limiter to restrict abuse of preview endpoints without affecting legitimate builder usage.

- Applied rate limiting to preview session creation (`startChatPreview`)
- Uses @upstash/ratelimit with a limit of 20 requests per minute
- Rate limiting is applied per authenticated user ID, or per IP address as a fallback
- Throws a `TOO_MANY_REQUESTS` error when the limit is exceeded
- Designed to prevent automated abuse while preserving normal developer experience

---

### ⚠️ Note

This does not affect normal builder usage, as limits are high enough for manual testing but prevent high-frequency automated abuse.

---

### 🧪 Testing

- Run local environment (`docker compose up -d` + `bun run dev`)
- Open preview repeatedly or hit `/preview/startChat` endpoint directly
- After ~20 requests/minute, API returns `TOO_MANY_REQUESTS`

---

### 📌 Future Improvements

- Extend rate limiting to `/sessions/continueChat` to fully prevent bypass via existing preview sessions
- Make limits configurable per environment


baptisteArno:
Feel free to re-open if you're willing to reply to Copilot suggestions :)

PR #2469, opened 2026-04-20
fix(security): strip Basic Auth credentials from HTTP request logs and sanitize upload filenames

beejak:
## Summary

Two security issues found in the HTTP request block execution path and file upload URL generation.

### 1. Basic Auth credentials leaked in execution logs (High)

`executeHttpRequestBlock.ts` extracts `username` and `password` from the `Authorization: Basic ...` header into a `basicAuth` object and spreads it onto the `request` object. That same `request` object is then serialized verbatim into every log entry — success, HTTP error, timeout, and unknown error.

Anyone with access to bot execution logs (the bot owner, workspace admins, or anyone with log storage access) can read plaintext passwords from any HTTP block that uses Basic Auth.

**Before:**
```ts
logs.push({
  status: "success",
  details: JSON.stringify({ statusCode: response.status, response: body, request }),
  //                                                                       ^ includes username + password
});
```

**After:**
```ts
const safeRequest = omit(request, "username", "password");
// safeRequest used in all logs; original request used for the actual call
```

### 2. Unsanitized filename in S3 key construction (Medium)

`handleGenerateUploadUrl.ts` incorporates the user-supplied `fileName` directly into the S3 object key and the `fileUrl` returned to the client without any sanitization. While S3 doesn't resolve `../` as filesystem traversal, the raw filename could contain path separators that produce unexpected key structures or confuse downstream URL parsing.

**Fix:** `basename(fileName)` applied before building the key, stripping any directory components.

## Test plan

- [ ] HTTP block with Basic Auth: execution logs should show `headers` and `url` but **no** `username`/`password` fields
- [ ] File upload with filename `../../../evil.txt` — S3 key and returned `fileUrl` should use `evil.txt` only
- [ ] Normal HTTP block execution and file upload continue to work as before

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/blocks/fileInput/src/api/handleGenerateUploadUrl.ts
+++ b/packages/blocks/fileInput/src/api/handleGenerateUploadUrl.ts
@@ -10,6 +10,7 @@ import { parseAllowedFileTypesMetadata } from "@typebot.io/lib/extensionFromMime
 import { generatePresignedPutUrl } from "@typebot.io/lib/s3/generatePresignedPutUrl";
 import prisma from "@typebot.io/prisma";
 import type { Prisma } from "@typebot.io/prisma/types";
+import { basename } from "path";
 import { z } from "zod";
 
 export const generateUploadUrlInputSchema = z.object({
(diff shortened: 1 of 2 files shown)


baptisteArno:
@codex review

baptisteArno:
Feel free to re-open if you are considering working more it.

PR #2474, opened 2026-04-27
🔒️ Add SSRF_ALLOWED_HOSTS env for self-hosted internal APIs

klinux:
## Summary

Self-hosted deployments often have legitimate internal corporate APIs on RFC1918 ranges (10/8, 172.16/12, 192.168/16) — e.g., a backend chat API exposed only on the internal cluster network. Since v3.14, the SSRF mitigation introduced for [CVE-2025-64709 / GHSA-8gq9-rw7v-3jpr](https://github.com/baptisteArno/typebot.io/security/advisories/GHSA-8gq9-rw7v-3jpr) blocks every private range unconditionally, which prevents HTTP Request blocks (and Function blocks via fetch) from reaching those APIs without exposing them to the public internet.

The advisory itself listed hostname allowlisting as one of the recommended mitigations (item #5: "Implement an SSRF-safe proxy or apply hostname allowlists for outgoing requests"), and this PR implements it as an opt-in env var.

## What changes

- New env var `SSRF_ALLOWED_HOSTS` (comma-separated hostnames) parsed in `packages/env`
- `validateHttpReqUrl` now accepts an `allowedHosts` parameter (symmetric with the existing `lookupHost` injection point); the env var is the default
- When the URL's hostname matches an entry, `validateIPAddress` is called with `{ allowPrivateRanges: true }`, which **only** skips the RFC1918 range checks (10/8, 172.16/12, 192.168/16)

## What the allowlist does NOT relax

Every other protection remains active even for allowlisted hosts:

- ✅ Link-local 169.254.0.0/16 — **the actual CVE vector** (AWS/GCP/Azure metadata)
- ✅ Loopback 127.0.0.0/8 and IPv6 ::1
- ✅ 0.0.0.0/8
- ✅ IPv6 link-local fe80::/10 and unique local fc00::/7
- ✅ Cloud metadata hostnames (\`metadata.google.internal\`, \`metadata.goog\`, \`metadata\`)
- ✅ \`localhost\` in production
- ✅ Decimal/hex/octal IP encoding bypasses
- ✅ IMDS bypass headers (\`X-aws-ec2-metadata-token*\`, \`Metadata-Flavor\`)

This is the deliberate design: **even if an attacker controls DNS for an allowlisted hostname and points it to 169.254.169.254, the link-local check still fires.** The allowlist intentionally narrows what's relaxed — corp LAN access, not metadata-service access.

## Test plan

- [x] All existing 53 SSRF tests still pass unchanged (default behavior preserved when env unset)
- [x] New \`describe\` block covering 14 cases:
  - RFC1918 hostnames pass when listed (10/8, 172.16/12, 192.168/16, direct IP literal)
  - Link-local **still blocks** for allowlisted host (DNS hijack defense)
  - Loopback **still blocks** for allowlisted host
  - Direct \`169.254.169.254\` IP literal **still blocks** even when listed
  - \`metadata.google.internal\` **still blocks** even when listed
  - Decimal-encoded metadata IP **still blocks** even when listed
  - Default behavior preserved when \`allowedHosts\` is undefined or empty
  - Hostname not in allowlist still blocks
  - Case-insensitive matching (URL parser normalizes hostname)
  - No subdomain wildcarding (exact match only)
- [x] \`bun test\` green: 63/63 in \`validateHttpReqUrl.test.ts\`
- [x] \`tsc --noEmit\` green for \`packages/lib\` and \`packages/env\`
- [x] Full \`nx affected\` test suite green (whatsapp, feature-flags, spaces, rich-text, root, emails, bot-engine, results, builder, lib — all passed)

## Use case

Currently, self-hosters facing this hit dead-ends: their internal corp DNS resolves to 10.x, the validator rejects it, and the only escape valves are (a) expose the API publicly (security regression — adds attack surface), (b) downgrade to ≤ v3.13.x (re-introduces the vulnerable code path), or (c) maintain a fork with the validator patched (fragile, breaks on every upgrade). An opt-in env var resolves this without weakening the core mitigation.

I'm opening a companion issue (#2475) explaining the use case in more detail and to gather feedback if a different design is preferred — happy to iterate.

### DIFF
--- a/packages/env/src/index.ts
+++ b/packages/env/src/index.ts
@@ -85,6 +85,16 @@ const baseEnv = {
       )
       .default("FREE"),
     TYPEBOT_DEBUG: boolean.optional().default(false),
+    SSRF_ALLOWED_HOSTS: z
+      .string()
+      .min(1)
+      .transform((val) =>
+        val
+          .split(",")
+          .map((s) => s.trim().toLowerCase())
+          .filter(Boolean),
+      )
+      .optional(),
     CHAT_API_TIMEOUT: z.coerce.number().optional(),
     RADAR_HIGH_RISK_KEYWORDS: z
       .string()
--- a/packages/lib/src/ssrf/validateHttpReqUrl.test.ts
+++ b/packages/lib/src/ssrf/validateHttpReqUrl.test.ts
@@ -27,6 +27,26 @@ const lookupHost = async (
     return [{ address: "10.0.0.5", family: 4 }];
   }
 
+  if (hostname === "internal.corp.example") {
+    return [{ address: "10.5.5.5", family: 4 }];
+  }
+
+  if (hostname === "172.corp.example") {
+    return [{ address: "172.20.10.5", family: 4 }];
+  }
+
+  if (hostname === "192.corp.example") {
+    return [{ address: "192.168.1.10", family: 4 }];
+  }
+
+  if (hostname === "hijacked.allowlisted.example") {
+    return [{ address: "169.254.169.254", family: 4 }];
+  }
+
+  if (hostname === "hijacked-loopback.example") {
+    return [{ address: "127.0.0.1", family: 4 }];
+  }
+
   throw new Error(`Unexpected hostname lookup: ${hostname}`);
 };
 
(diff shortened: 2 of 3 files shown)


baptisteArno:
@codex review

baptisteArno (approved):

PR #2476, opened 2026-04-28
fix the button position close in modal

martinsjr:
When clicking on an image in the chat preview/runtime, the X button to close the modal appeared misplaced outside the correct area. This happened because the button was positioned with `position: fixed`, remaining stuck to the entire page viewport instead of being inside the modal/chat container.

before:
image

after:
image


baptisteArno (approved):

martinsjr:
I'm happy to help, even if it's just a small detail.

PR #2502, opened 2026-05-21
🐛 Fix unsafe upload URL generation

baptisteArno:
- Replace direct browser presigned PUT uploads with signed Typebot upload proxy URLs.
- Generate or validate upload object keys server-side while preserving legacy v1/v2/v3 file-input upload contracts.
- Keep builder slot uploads stable for replaceable assets and use generated names for runtime file uploads.
- Store active file-input MIME types as safe attachment downloads while keeping safe image uploads inline.
- Update upload clients and docs to support both raw PUT proxy uploads and form-data uploads.


baptisteArno:
@codex review

PR #2504, opened 2026-05-22
fix: include cookies in HTTP request executed from client

### LINKED ISSUE #1868: Add option to include cookies in HTTP request executed from client

dashitongzhi:
This change adds `credentials: include` to the fetch call when executing HTTP requests from the client (non-preview mode). This allows cookies to be properly sent and stored by the browser when making HTTP requests.

Fixes #1868

### DIFF
--- a/packages/embeds/js/src/features/blocks/integrations/httpRequest/executeHttpRequest.ts
+++ b/packages/embeds/js/src/features/blocks/integrations/httpRequest/executeHttpRequest.ts
@@ -10,7 +10,7 @@ export const executeHttpRequest = async (
       method,
       body: method !== "GET" && body ? JSON.stringify(body) : undefined,
       headers,
-      credentials: isPreview ? "omit" : undefined,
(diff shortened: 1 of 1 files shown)


baptisteArno:
This was intentional for security reasons

PR #2511, opened 2026-05-24
🐛 Block IPv6 unspecified SSRF targets

baptisteArno:
- Block IPv6 unspecified addresses in the shared SSRF IP validator.
- Add validator regressions for compressed and expanded IPv6 unspecified literals.
- Add a safeKy regression that verifies [::] requests are rejected before reaching a local IPv6 wildcard listener.


sondt99:
Thank you! @baptisteArno please publish this advisory https://github.com/baptisteArno/typebot.io/security/advisories/GHSA-qx46-p88f-xxm3

PR #2525, opened 2026-06-12
👌 Add WhatsApp webhook forwarding options

baptisteArno:
- Adds a WhatsApp deploy setting to enable webhook forwarding and choose between marketing/failed statuses, all statuses, or full Meta webhook payloads.
- Adds settings schema/constants for the new forwarding config while keeping legacy errorAndMarketingStatusWebhookForwardUrl fallback behavior.
- Updates production forwarding to route by scope and preserve the raw Meta payload for the full-payload option.
- Updates the WhatsApp status forwarding script to read/write the new config shape.
- Adds unit coverage for legacy fallback, scope precedence, disabled bots, duplicate URLs, filtered statuses, and full-payload forwarding.


baptisteArno:
@codex review

baptisteArno:
Needs rework.

PR #2531, opened 2026-06-17
🐛 Fix stored WhatsApp phone lookup fallback

baptisteArno:
- Fallback to the saved WhatsApp credential name when Meta phone lookup fails for an existing stored credential.
- Keep new Meta token setup validation strict so setup errors still surface.


baptisteArno:
@codex review

PR #2540, opened 2026-06-27
🐛 Fix Sentry quick-win errors

baptisteArno:
- Return a no-op response for Stripe subscription updates when the customer is no longer tied to a workspace.
- Validate user profile string fields against database limits before Prisma updates.
- Return a bad request for imported typebots missing the legacy start block or group.
- Truncate saved answer content to the MySQL TEXT byte limit before persisting.
- Cap WhatsApp interactive body text at Meta's 1024-character limit and add conversion tests.


baptisteArno:
@codex review

PR #2543, opened 2026-06-27
🐛 Fix Send Email attachment handling

baptisteArno:
- Restrict Send Email attachments to valid Typebot upload URLs.
- Resolve private Typebot upload URLs to short-lived storage URLs before sending.
- Enable Nodemailer local file access protection for Send Email messages.
- Fix WhatsApp public media upload keys to avoid a duplicated public prefix.
- Add regression tests for rejected attachment origins and valid Typebot uploads.


baptisteArno:
@codex review

PR #2550, opened 2026-06-30
fix: restore automatic database sync in dev

KavitaR45:
## Summary

Restores automatic Prisma client generation and database schema synchronization before local development servers start.

## Root cause

Before the Nx migration, the Turbo `dev` task depended on `^db:generate` and `^db:push`. Those dependencies were lost during the migration, leaving fresh local PostgreSQL databases without the required schema.

## Changes

- Add Prisma generation and schema push as defaults for Nx `dev` targets.
- Preserve those dependencies explicitly for builder and viewer, whose project-level `dependsOn` configuration overrides target defaults.
- Keep the local installation guide unchanged because `bun dev` once again initializes the database automatically.

## Validation

- Verified the generated Nx task graph: builder, viewer, and workflows all depend on the deduplicated Prisma `db:generate` and `db:push` tasks.
- Ran `bunx nx format-and-lint`.
- Passed the full affected pre-commit suite: formatting, repository lint, broken-link checks, and tests.

### DIFF
--- a/apps/builder/package.json
+++ b/apps/builder/package.json
@@ -8,6 +8,8 @@
     "targets": {
       "dev": {
         "dependsOn": [
+          "^db:generate",
+          "^db:push",
           {
             "projects": [
               "@typebot.io/react"
--- a/apps/docs/contribute/guides/local-installation.mdx
+++ b/apps/docs/contribute/guides/local-installation.mdx
@@ -36,7 +36,17 @@ icon: "laptop"
 
    This will also create a minio instance to enable file uploads.
 
-5. Start the builder and viewer
+5. Initialize the database
+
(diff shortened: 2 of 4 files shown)


baptisteArno:
Thanks for reporting this! You were right that a fresh local database was no longer initialized.

The database sync used to run automatically, but that dependency was accidentally dropped during our Nx migration. I updated your PR to restore the automatic `db:generate` and `db:push` tasks before the dev servers start, and removed the manual setup step.

Thanks for surfacing the regression!

baptisteArno (approved):
The Nx task graph now restores the automatic Prisma generation and schema push before the dev servers start. Typecheck and the affected local checks pass.

PR #2553, opened 2026-07-03
🔧 Restore OpenAPI docs generation pipeline

baptisteArno:
## Summary

- Restores OpenAPI docs generation, lost in the oRPC migration: new `generate-openapi` nx targets regenerate `apps/docs/openapi/{builder,viewer}.json` offline from the oRPC routers, reusing the exact options served by the live `/api/openapi.json` endpoints. Both specs are regenerated in this PR (builder.json was stale since March).
- Fixes spec generation crashing on the spaces router (Effect Schema had no JSON Schema converter — this also breaks the live builder `/api/openapi.json` today) and sanitizes an `AppendValue(s)` component name that made Mintlify silently render every builder API page empty.
- Keeps the docs from going stale again: pre-commit regenerates affected specs and auto-stages them, CI fails PRs with stale specs, and a new check requires lockstep `@typebot.io/js` / `@typebot.io/react` version bumps whenever embed files change.

## Test plan

- [x] `bunx nx generate-openapi builder viewer` produces valid OpenAPI 3.1 files matching the live endpoints (works without env: isolated-vm stubbed, dummy DATABASE_URL auto-matched to the generated Prisma client)
- [x] All 34 `openapi:` references in `apps/docs/api-reference/*.mdx` resolve against the new specs; pages verified rendering in `mintlify dev` (params, unions, curl samples)
- [x] Embeds version check: fails when js or react files change without both bumps, passes on the historical 0.10.5 bump commit, ignores unrelated changes
- [x] Pre-commit pipeline (lint, tests, generation, embeds check) passes end to end
- [ ] Confirm the "Check OpenAPI docs are up to date" step passes on this PR


baptisteArno:
@codex review

baptisteArno:
Addressed the review in 7c4293a:

- **Regenerate specs even when only generated docs change**: the CI gate now runs `nx run-many -t generate-openapi` unconditionally before diffing, so manual edits to the generated files are rejected.
- **Compare embed versions instead of matching added lines**: the script now parses the version from the merge base and the working tree and requires them to differ.
- **Preserve operationIds**: the seven operations that existed in the previously published viewer spec (`startChat`, `continueChat`, `saveClientLogs`, `startChatPreview`, `updateTypebotInSession`, `generateUploadUrl`, `whatsAppRouter-subscribeWebhook`) now declare their legacy `operationId` in route metadata, so the live spec and docs keep stable SDK method names.
- **Keep documented error responses**: standard `400/401/403/500` errors are now declared on both oRPC base builders, so every operation documents them again (matching the previous specs).
- **`viewer` MDX qualifier**: verified working — the spec basename is accepted by Mintlify (same form as the pre-existing `openapi: builder GET ...` in `get-stats.mdx`), and the three chat pages render the correct viewer operation with `https://typebot.io/api` servers in `mintlify dev`. Left as is.

@codex review

baptisteArno:
Addressed in bf469aa:

- **Preserve builder operationIds**: all 48 operations that exist in both the previously published builder spec and the current API now pin their legacy `operationId` in route metadata (the remaining 6 legacy ids belong to removed billing subscription endpoints).
- **Keep the published auth scheme name**: the scheme is `Authorization` again in both specs and all security references.
- **Document thrown ORPC error codes**: added `NOT_FOUND` to both base error maps, restoring the documented `404` responses. `CONFLICT`/`TOO_MANY_REQUESTS` were never documented in the previously published specs (they only documented 200/400/401/403/404/500), so declaring them is deliberately left for a follow-up where they can be declared on the specific procedures that throw them.
- **Require the embed version to increase**: the check now compares semver order against the merge base, not mere inequality.
- **Empty DELETE request bodies**: leaving as generated. The committed files intentionally mirror what the live `/api/openapi.json` endpoints serve, and this optional-empty-body shape is the oRPC generator's output for compact-input routes; suppressing it only in the committed files would desync docs from the live spec. Happy to revisit upstream or with `inputStructure: "detailed"` per route if it causes real client issues.

@codex review

baptisteArno:
Addressed in the latest commit:

- **Enforce equal embed package versions**: after both bumps are validated, the script now also requires `@typebot.io/js` and `@typebot.io/react` to be set to the same version.
- **Compare embed bumps to the PR base**: changed files are still detected against the merge base (so target-branch drift does not false-trigger the check), but versions are now compared against the base ref tip, so a bump that races an already-merged release fails.
- **OpenAPI frontmatter**: switched the three chat pages to the documented `/openapi/viewer.json POST ...` file-path form and verified they render the correct viewer operations in `mintlify dev`.

@codex review

baptisteArno:
Wrapping up the review loop — the two remaining suggestions are tracked as follow-ups rather than addressed here:

- **Filtering internal routes from the docs**: the committed specs intentionally mirror the live `/api/openapi.json` output. Curating a public-only subset (e.g. an opt-out flag in route metadata) is a docs-scoping decision to make separately.
- **Declaring CONFLICT/TOO_MANY_REQUESTS/TIMEOUT**: never documented in the previously published specs; the precise fix is per-procedure error declarations where those codes are actually thrown.

PR #2573, opened 2026-08-02
👌 Add a separator label for date range answers

### LINKED ISSUE #771: Translation for "to" inside a date message
### Discussed in https://github.com/baptisteArno/typebot.io/discussions/768

<div type='discussions-op-text'>

<sup>Originally posted by **jmchat** September  5, 2023</sup>
I can't find the option to change the "to" word when you enter a start and enddate. I can change the text inside the datepicker but not when it's placed after you are done with the selection, so I got this "11-09-2023 to 14-09-2023" Where I want to change it to "11-09-2023 tot 14-09-2023" (which is Dutch) or "11-09-2023 - 14-09-2023" is also fine.

Am I missing something or is this not possible?</div>

sridhar-3009:
Closes #771.

The word joining the start and end date in a range answer ("11-09-2023 **to** 14-09-2023") was hardcoded in two places:
- `parseDateReply.ts`, which builds the value shown when someone types a free-text reply (e.g. on WhatsApp)
- `DateForm.tsx`, the web widget's date picker form

Neither used the block's `labels` option, so there was no way to change it. The existing `labels.to` field doesn't cover this — it labels the "To" field above the second date input in the form, which is a different string ("To:" vs the lowercase joining word), so I added a separate `labels.separator` field instead of overloading it.

Now `labels.separator` (default `"to"`) drives the joining word in both places, and shows up as a new field in the block's settings panel in the builder, right under the existing From/To label fields.

Added a few unit tests for `parseDateReply` covering the default separator, a custom one, and the non-range case.

### DIFF
--- a/apps/builder/src/features/blocks/inputs/date/components/DateInputSettings.tsx
+++ b/apps/builder/src/features/blocks/inputs/date/components/DateInputSettings.tsx
@@ -19,6 +19,8 @@ export const DateInputSettings = ({ options, onOptionsChange }: Props) => {
     onOptionsChange({ ...options, labels: { ...options?.labels, from } });
   const updateToLabel = (to: string) =>
     onOptionsChange({ ...options, labels: { ...options?.labels, to } });
+  const updateSeparatorLabel = (separator: string) =>
+    onOptionsChange({ ...options, labels: { ...options?.labels, separator } });
   const updateButtonLabel = (button: string) =>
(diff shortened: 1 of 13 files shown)


baptisteArno:
Thanks a lot for the contribution @sridhar-3009! I pushed a small follow-up to keep the date picker payload compatible with the server parser and resolve variables in the separator. Everything looks good now 🙏

PR #2574, opened 2026-08-02
Allow any file type in Embed bubble to be sent as a WhatsApp document

### LINKED ISSUE #240: Allow for any file type URL in Embed bubble to display them as downloadable on WhatsApp

sridhar-3009:
Closes #240.

`convertMessageToWhatsAppMessage` only sends an Embed bubble as a downloadable WhatsApp "document" when its URL's extension exactly matches one of the ~15 non-media entries in `extensionFromMimeType` (pdf, doc/docx, xls/xlsx, ppt/pptx, zip, rar, 7z, json, xml, js, tar, gz, apk, exe, mpkg, rtf, sh, ttf, odt/ods/odp). Anything else — .csv, .txt, .epub, or any format not on that list — fell through to a plain text link instead.

## Fix

Flipped the check: instead of requiring the extension to be in that whitelist, it now only excludes known image/audio/video extensions (derived from the same `extensionFromMimeType` map) — any other file extension is sent as a document, matching the ask ("any file type").

While in there, I also fixed how the extension/filename get derived: the old code did `url.split(".").pop()`, which for a bare domain like `https://example.com` (no path) would grab "com" as a fake extension. Parsing with `new URL(...)` and reading the pathname's last segment avoids that, and also naturally strips query params/hash (a signed URL like `file.pdf?token=...` now correctly extracts `.pdf`, which the old code didn't handle either).

## Tests

Added `convertMessageToWhatsAppMessage.test.ts` (this function had no test coverage before) covering:
- a known document extension (.pdf) → document
- an extension not in the whitelist (.csv) → now also a document
- query params stripped before extracting the extension
- an image extension (.png) → still a plain text link, not a document
- a bare domain isn't mistaken for a filename
- a path with no dot → plain text link
- no url → null

All 53 existing + new tests in the whatsapp package pass.

### DIFF
--- a/packages/whatsapp/src/convertMessageToWhatsAppMessage.test.ts
+++ b/packages/whatsapp/src/convertMessageToWhatsAppMessage.test.ts
@@ -1,0 +1,99 @@
+import { BubbleBlockType } from "@typebot.io/blocks-bubbles/constants";
+import type { ContinueChatResponse } from "@typebot.io/chat-api/schemas";
+import { describe, expect, it } from "vitest";
+import { convertMessageToWhatsAppMessage } from "./convertMessageToWhatsAppMessage";
+
+const embedMessage = (url: string): ContinueChatResponse["messages"][number] =>
+  ({
+    type: BubbleBlockType.EMBED,
+    content: { url },
+  }) as ContinueChatResponse["messages"][number];
+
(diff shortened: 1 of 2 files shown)


baptisteArno:
Allowing any file type is a security issue. We’d rather keep an explicit allowlist of supported file types.

PR #2575, opened 2026-08-02
Add option to include cookies in HTTP request executed from client

### LINKED ISSUE #1868: Add option to include cookies in HTTP request executed from client

sridhar-3009:
Closes #1868.

For an HTTP Request block with "Execute on client" enabled, the browser `fetch` call had no `credentials` option set for non-preview requests, so it fell back to the browser default (`same-origin`). A cookie set by a cross-origin server in the response body was therefore never stored by the browser — there was no way to opt into cross-origin cookie handling.

## Fix

Added a `withCredentials` option to the HTTP Request block, only shown in the builder when "Execute on client" is enabled:
- `packages/blocks/integrations/src/httpRequest/schema.ts`: added to both `httpRequestOptionsV5Schema` (the block option) and `executableHttpRequestSchema` (the payload sent to the client)
- `executeHttpRequestBlock.ts`: threads `block.options?.withCredentials` through `parseHttpRequestAttributes` into the client-side action payload
- `executeHttpRequest.ts` (client): `credentials: isPreview ? "omit" : withCredentials ? "include" : undefined` — preview still always omits credentials, unchanged
- Added a "Send cookies" toggle next to "Execute on client" in `HttpRequestAdvancedConfigForm.tsx`

## Tests

Added `executeHttpRequest.test.ts` (no test coverage existed for this function before) covering: default (no credentials option), `withCredentials: true` → `"include"`, and preview mode still omitting credentials even when `withCredentials` is true.

Ran typecheck across all affected projects (`nx affected -t typecheck`) clean, and regenerated the OpenAPI docs (the new field shows up wherever `HttpRequestBlock`'s options are referenced — HTTP Request, Zapier, Make.com, Pabbly Connect all share the schema).

### DIFF
--- a/apps/builder/src/features/blocks/integrations/httpRequest/components/HttpRequestAdvancedConfigForm.tsx
+++ b/apps/builder/src/features/blocks/integrations/httpRequest/components/HttpRequestAdvancedConfigForm.tsx
@@ -113,6 +113,9 @@ export const HttpRequestAdvancedConfigForm = ({
   const updateIsExecutedOnClient = (isExecutedOnClient: boolean) =>
     onOptionsChange({ ...options, isExecutedOnClient });
 
+  const updateWithCredentials = (withCredentials: boolean) =>
+    onOptionsChange({ ...options, withCredentials });
+
   const ResponseMappingInputs = useMemo(
     () =>
       function Component(props: TableListItemProps<ResponseVariableMapping>) {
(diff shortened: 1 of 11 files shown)


baptisteArno:
Thanks for working on this! We intentionally changed this behavior for security reasons, so we don't want to add an option that re-enables sending cookies with client-side HTTP requests. I'm closing this PR.

PR #2579, opened 2026-08-08
docs: add You Secure (Brazil) self-hosting deploy guide

gabrielkemmer:
Adds a self-hosting deploy guide for You Secure, a VPS provider with infrastructure in São Paulo, Brazil.

Motivation: Typebot has a sizeable Brazilian user base, and the existing self-hosting guides all assume infrastructure outside the country. For teams serving Brazilian users this matters twice — round-trip latency, and keeping the Postgres database that holds typebot results inside Brazil.

The guide follows the same shape as the existing alibaba-cloud page: DNS setup for the builder/viewer hostnames, the deploy step, configuration pointers back to the existing docs, upgrade, and backups. It links back to ./docker for anyone who'd rather do it by hand, and keeps the standard cloud note at the top.

Also registers the page in apps/docs/mint.json under the self-hosting Deploy group.

I'm affiliated with You Secure — happy to adjust the tone, trim it, or drop it entirely if you'd rather keep the provider list short.

### DIFF
--- a/apps/docs/mint.json
+++ b/apps/docs/mint.json
@@ -257,6 +257,7 @@
         "self-hosting/deploy/docker",
         "self-hosting/deploy/vercel",
         "self-hosting/deploy/alibaba-cloud",
+        "self-hosting/deploy/you-secure",
         "self-hosting/deploy/manual"
       ]
     },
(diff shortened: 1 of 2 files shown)


baptisteArno:
Thanks for the contribution! This deployment option is too niche to include in the official docs, so I’m closing this PR.

PR #2585, opened 2026-08-19
Fix date and time limit validation bypass (#2581)

### LINKED ISSUE #2581: [BUG] When you set limits for a date + time and the user does not choose the date but enters the date himself, then the limits can be circumvented.
**Describe the bug**
When you set limits for a date + time and the user does not choose the date but enters the date himself, then the limits can be circumvented.

**Expected behavior**
An error message that is shown just like the error message when a wrong email is being entered in a text field, to stay within the boundaries.

harshdhankhar12:
Fixes #2581

### Bug Description
When setting `min` / `max` limits for a date + time input, users could circumvent limits by manually typing dates into the input field rather than picking them from the calendar UI.

### Root Cause
1. **Frontend**: `DateForm.tsx` called `e.preventDefault()` on form submit without checking `form.checkValidity()`, bypassing HTML5 `min` / `max` constraint validation.
2. **Backend**: `parseDateReply.ts` parsed user replies using `parseDateWithNeutralTimezone` but evaluated `min` / `max` with `new Date(limit)`, creating timezone offset discrepancies.

### Solution
1. **Frontend**: Attached `formRef` to the date input form and added `formRef.checkValidity()` / `formRef.reportValidity()` before submitting. Out-of-bounds typed values now trigger the browser validation tooltip and block submission.
2. **Backend**: Wrapped `new Date(max)` and `new Date(min)` in `parseDateWithNeutralTimezone` to align boundary checks on the exact same timezone scale.

### DIFF
--- a/packages/bot-engine/src/blocks/inputs/date/parseDateReply.spec.ts
+++ b/packages/bot-engine/src/blocks/inputs/date/parseDateReply.spec.ts
@@ -1,0 +1,142 @@
+import { InputBlockType } from "@typebot.io/blocks-inputs/constants";
+import type { DateInputBlock } from "@typebot.io/blocks-inputs/date/schema";
+import type { SessionStore } from "@typebot.io/runtime-session-store";
+import type { Variable } from "@typebot.io/variables/schemas";
+import { describe, expect, it } from "vitest";
+import { parseDateReply } from "./parseDateReply";
+
+const createDateBlock = (
+  options?: DateInputBlock["options"],
+): DateInputBlock => ({
+  id: "block-1",
+  type: InputBlockType.DATE,
+  options,
+});
+
+describe("parseDateReply", () => {
+  it("parses valid date replies", () => {
+    const block = createDateBlock();
+    const result = parseDateReply("2024-05-05", block);
+    expect(result.status).toBe("success");
+    if (result.status === "success") {
+      expect(result.content).toBe("2024-05-05");
+    }
+  });
+
+  it("parses valid datetime replies when hasTime is enabled", () => {
+    const block = createDateBlock({ hasTime: true });
+    const result = parseDateReply("2024-05-05 14:30", block);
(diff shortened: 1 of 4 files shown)


baptisteArno:
Thanks for working on this. Please fix these before we can merge it:

- Rebase the branch so it only contains the date/time change. The custom-head commit belongs in #2584.
- Fix the non-UTC server regression in `parseDateReply`. In `Europe/Paris`, a 09:00 reply can pass a 10:00 minimum, while valid in-range values can fail. `America/Los_Angeles` shows the inverse around the maximum.
- Resolve variable-backed min/max validation and the exact-date maximum boundary.
- Add focused tests for manual date/datetime entry, ranges, min/max boundaries, variables, and at least UTC plus one non-UTC timezone.

Ping me when it is updated and I will take another look.

harshdhankhar12:
Hi @baptisteArno , thanks for the thorough review! I've updated the PR with all requested fixes:

1. **Clean Rebase**: Rebased the branch onto `main` so it strictly contains only date/time changes (removing the custom-head commit belonging to #2584).
2. **Server Timezone Independence**: Updated `parseDateReply.ts` to compare date/time components using deterministic `Date.UTC(...)` timestamps, removing dependence on `getTimezoneOffset()` of the Node server.
3. **Exact-Date Maximum Boundary**: Handled date-only `max` boundaries at `23:59:59.999` so exact dates pass validation.
4. **Variable-Backed Limits**: Passed `{ variables, sessionStore }` into `parseDateReply` to evaluate variable-backed `min`/`max` bounds.
5. **Unit Test Suite**: Created `parseDateReply.spec.ts` testing manual entry, ranges, boundaries, variables, and timezone independence across `UTC`, `Europe/Paris`, and `America/Los_Angeles`.

Ready for another look!

harshdhankhar12:
Hi @baptisteArno  I've updated the PR with all requested fixes:

PR #2603, opened 2026-09-14
🐛 Block external sign-in redirects

baptisteArno:
Prevent post-login navigation outside the builder by accepting only root-relative redirect paths. SignInForm, the `/typebots` proxy and email magic-link callbacks now share the same guard; internal query parameters and fragments are preserved, including nested callback returns.

Reject absolute and protocol-relative URLs, backslashes, controls, and dot-segment normalization that would expose an authority. Absolute same-origin URLs and bare relative paths now fall back to `/typebots`. Encoded path data is not decoded again. The proxy uses the destination query rather than carrying unrelated callback parameters forward.

Validation:
- 40 focused unit regressions and 27 Chromium scenarios pass, covering existing sessions, synthetic Auth.js sign-in and proxy returns with the actual Next 16.2.9 router.
- Manual Helium checks confirm the external fallback and internal query/hash return.
- Repository pre-commit checks passed (affected lint, tests, builds, links and OpenAPI generation); auth typecheck passed. The earlier standalone builder typecheck reported errors outside this diff, including oRPC and SpacesList.
- `tests/auth-redirect/README.md` documents the local preview and compatibility limits. The fixture stubs presentation and uses a synthetic local identity; real email delivery, external OAuth providers, Firefox and WebKit were not tested.


baptisteArno:
@codex review

baptisteArno:
@codex review

